PR #4, opened 2017-06-20
Two small fixes

nuance:
I encountered a few small bugs when trying this out, these commits appear to fix both. The errors I encountered were:

- the `TokenTextEncoder` constructor choked when I provided it with a vocab filename, both in that it didn't read the file (which caused `id_to_token` to not be initialized), and the super class constructor was called after some code that relied on the `num_reserved_ids` property to be set.
- the `build_to_target_size` method on `SubwordTextEncoder` could invert min / max during the binary search if the target vocab size was larger than the actual vocab size, which triggered an infinite loop with min 1 / max -1. I'm not sure if this is the best fix (should it short circuit return if the source vocab is already smaller than the target?).


nuance:
CLA email should be updated.

PR #6, opened 2017-06-20
v1.0.3

rsepassi:
Push internal changes out.


rickyhan:
Hi do the algorithmic examples work now?

lukaszkaiser:
It generates the right data now, but one more correction is needed (in problem_hparams). Coming soon.

lukaszkaiser:
Just change "algorithmic_reverse_decimal40": lambda p: algorithmic(11, p) to ...[the same] algorithmic(12, p). Re-generate the data and re-train, I'm just doing that, will check if that's the last bug.

Issue #11, opened 2017-06-21
Is GPU requirement a requirement?


skeydan:
Hi,

I'd very much like to try this, but I don't have an nvidia gpu... Is the dependency on tensorflow-gpu a hard requirement?

Thanks a lot
Sigrid

odunboye:
will also like to know

lukaszkaiser:
It shouldn't be, I can run tensor2tensor on CPU only. Does the pip package require GPU? If so, we should change it. I think you should be able to pip install tensorflow (no gpu) and then pip install tensor2tensor, does that work?

skeydan:
Hi, thanks!
Yes, you have the requirement in setup.py... I already have cpu-only tensorflow installed, but when I run pip install tensor2tensor it doesn't want that one and crashes with the libaCUDA... thing...

stefan-it:
I can also confirm that `tensor2tensor` is working on a CPU-only computer. 

@skeydan Could you try running the `pip` installation with `--no-dependencies`? So it hopefully won't install the gpu version of *TensorFlow* :)

rsepassi:
1.0.5 does not require tensorflow-gpu.

You can install tensorflow (cpu or gpu versions) separately or have it with tensor2tensor:

```
pip install tensor2tensor # assumes tensorflow already installed
pip install tensor2tensor[tensorflow] # will install cpu version
pip install tensor2tensor[tensorflow_gpu] # will install gpu version
```

PR #12, opened 2017-06-22
setuptools: do not use gpu version of TensorFlow as hard requirement

stefan-it:
Hi,

as discussed in #11 the gpu version of *TensorFlow* is not a hard requirement for `tensor2tensor`, as it can be run on cpu-only.

This PR removes the dependency for `tensorflow_gpu` in `setup.py`.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -17,7 +17,7 @@ setup(
         'numpy',
         'sympy',
         'six',
-        'tensorflow-gpu>=1.2.0rc1',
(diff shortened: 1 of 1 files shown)


stefan-it:
I signed it!

rsepassi:
Thanks @stefan-it! We've actually just addressed this issue in the latest push. 1.0.5 no longer requires tensorflow-gpu. Give it a few minutes to show up on PyPI.

Issue #2, opened 2017-06-18
Python3 compatibility


vthorsteinsson:
There are some holes in the Python 3 compatibility of the Tensor2tensor code. For instance:

In data_generators/generator_utils.py, `import urllib` needs to be:

```
import sys
if sys.version_info[0] >= 3:
  import urllib.request as urllib
else:
  import urllib
```

In data_generators/image.py, `import cPickle` needs to be:

```
try:
  import cPickle
except ImportError:
  import pickle as cPickle

```
Finally, data_generators/tokenizer.py needs to be revised as it assumes that a char ordinal is always in the range (0, 256), which is not a safe assumption in Python 3. A better solution uses a set instead of array subscripts based on char ordinals. Would you like me to submit a revised version in a pull request?

rsepassi:
Thank you! Yes, please do. The chr assumption may also be in a few other places - e.g. text_encoder.py. Appreciate the help. We're on 2.7 at Google so Python 3 usage isn't as well-exercised/tested.

plutusedge:
Thanks for the info this is helpful

vthorsteinsson:
See PR #22

rsepassi:
PR #22 merged. Thanks!

PR #22, opened 2017-06-22
Python3 compatibility; better Unicode support

vthorsteinsson:
This PR includes a set of changes to allow the t2t-datagen program to run to completion under Python 3.x. It fixes imports of urllib and cPickle and removes the assumption that strings contain bytes. It also makes minor bug fixes and efficiency enhancements. Functionality-wise, it modifies the subword vocabulary builder so that **all** (Unicode) characters that occur in the source text are automatically included in the vocabulary, vs. the previous design that included all characters with ordinals 0..255 inclusive, regardless of whether they occurred in the vocabulary or not. Also, in the ByteTextEncoder, Unicode characters with ordinals >= 256 are replaced by a marker character (currently ×, chr(215)).

This fixes issue #2 

Caveat: the tests are not updated by this PR.


tensor2tensor/data_generators/generator_utils.py
```diff
@@ -22,12 +22,12 @@
 import io
 import os
 import tarfile
-import urllib
 
 # Dependency imports
 
 import six
 from six.moves import xrange  # pylint: disable=redefined-builtin
+import six.moves.urllib_request
```
rsepassi: Do uses need to be updated? `urllib->urllib_request`

tensor2tensor/data_generators/text_encoder.py
```diff
@@ -73,7 +78,7 @@ def decode(self, ids):
     decoded_ids = []
     for id_ in ids:
       if 0 <= id_ < self._num_reserved_ids:
-        decoded_ids.append(RESERVED_TOKENS[int(id_)])
```
rsepassi: should keep the int cast because the ids coming in may be numpy ints which cannot be used for indexing.

tensor2tensor/data_generators/text_encoder.py
```diff
@@ -36,6 +37,10 @@
 EOS = '<EOS>'
 RESERVED_TOKENS = [PAD, EOS]
 
+# Character that will be used instead of those with
+# ordinal numbers > 255
+UNK_CHAR = '×' # ordinal 215
```
rsepassi: why 215 for unk?

tensor2tensor/data_generators/text_encoder.py
```diff
@@ -113,18 +124,18 @@ def __init__(self, vocab_filename, reverse=False, num_reserved_ids=2):
     self._reverse = reverse
     if vocab_filename is not None:
```
rsepassi: here a vocab filename has to be provided so remove the line `if vocab_filename is not None:` and the `else` stanza.

tensor2tensor/data_generators/text_encoder.py
```diff
@@ -243,15 +254,22 @@ def _escaped_token_to_subtokens(self, escaped_token):
...
         end -= 1
       ret.append(subtoken)
-      pos = end
+      if end > pos:
+        pos = end
+      else:
+        # This kinda should not happen, but it does. Cop out by skipping the
+        # nonexistent subtoken from the returned list.
+        # print("Unable to find subtoken in string '{0}'".format(escaped_token))
+        pos += 1
```
rsepassi: why the change here? what's motivating it and what is this code intending?

tensor2tensor/data_generators/text_encoder.py
```diff
@@ -337,31 +355,33 @@ def build_from_token_counts(self,
             starts.append(pos)
             pos += len(self.subtoken_to_subtoken_string(subtoken))
         for start in starts:
-          for end in xrange(start + 1, len(escaped_token) + 1):
```
rsepassi: don't we need this to go 1 past the final `start`?

tensor2tensor/data_generators/text_encoder.py
```diff
@@ -337,31 +355,33 @@ def build_from_token_counts(self,
...
-            counts[subtoken_string] = counts.get(subtoken_string, 0) + count
+            counts[subtoken_string] += count
       # array of lists of candidate subtoken strings, by length
       len_to_subtoken_strings = []
       for subtoken_string, count in six.iteritems(counts):
-        if count < min_count or len(subtoken_string) <= 1:
+        lsub = len(subtoken_string)
+        # all subtoken strings of length 1 are included regardless of count
+        if count < min_count and lsub != 1:
           continue
```
rsepassi: is it possible to get empty subtokens? if so, we'll want to skip those as well.

tensor2tensor/data_generators/tokenizer.py
```diff
@@ -45,29 +45,24 @@
...
 from six.moves import xrange  # pylint: disable=redefined-builtin
-
+from collections import defaultdict
 
 class Tokenizer(object):
   """Vocab for breaking words into wordpieces.
   """
 
   def __init__(self):
     self._separator_chars = string.punctuation + string.whitespace
```
rsepassi: rm line

tensor2tensor/data_generators/tokenizer.py
```diff
@@ -45,29 +45,24 @@
...
-    self._separator_char_mask = array.array(
-        "l", [chr(i) in self._separator_chars for i in xrange(256)])
-    self.token_counts = dict()
+    self._separator_char_set = set(self._separator_chars)
```
rsepassi: `self._separator_char_set = set(string.punctuation + string.whitespace)`

tensor2tensor/data_generators/tokenizer.py
```diff
@@ -45,29 +45,24 @@
...
+    self.token_counts = defaultdict(int)
 
   def _increment_token_count(self, token):
```
rsepassi: remove this method and replace the 2 calls with `self.token_counts[token] += 1`

tensor2tensor/data_generators/generator_utils.py
vthorsteinsson: No. This is the official `six` way to import what used to be called `urllib` in Python2 but is called `urllib.request` in Python3 (or at least the part that is being used here, since urllib was split up into multiple sub-packages). No other changes are necessary as a result of this modified import.

tensor2tensor/data_generators/text_encoder.py
vthorsteinsson: Because it looks kinda like an "unknown" token and is unlikely to clash with significant characters. We have to do something with Unicode characters with ordinals > 255, and mapping them to an 'indicator' character instead of skipping them feels sensible.

tensor2tensor/data_generators/text_encoder.py
vthorsteinsson: OK, agreed.

tensor2tensor/data_generators/text_encoder.py
vthorsteinsson: OK - the if statement came from an earlier fix, see https://github.com/tensorflow/tensor2tensor/commit/2f03f0cad63c2142595aaffdacc72427496653f3

tensor2tensor/data_generators/text_encoder.py
vthorsteinsson: If the `escaped_token[pos:end]` substring is for some reason not found in `self._subtoken_string_to_id`, the code decrements `end` and continues. In degenerate cases (which do appear in the test data sets, at least when I run `t2t-datagen` on Python3), `end` may become <= `pos`, meaning that the `escaped_token` slice becomes an empty string, which is never found in `self._subtoken_string_to_id`. In this case, the `while True` loop does not terminate and the program hangs.

tensor2tensor/data_generators/text_encoder.py
vthorsteinsson: No, `end` is only used as a slice endpoint of `escaped_token` and you never need (or want) its value to go beyond `len(escaped_token)`, which is one past the last index of the string. The last character of the string is `escaped_token[len(escaped_token)-1 : len(escaped_token)]`.

tensor2tensor/data_generators/text_encoder.py
vthorsteinsson: No, that is not possible. But the code could be made more explicit (at the cost of becoming a bit uglier).

tensor2tensor/data_generators/tokenizer.py
vthorsteinsson: Agreed.

tensor2tensor/data_generators/tokenizer.py
vthorsteinsson: Don't understand?

tensor2tensor/data_generators/tokenizer.py
vthorsteinsson: Yup.

tensor2tensor/data_generators/text_encoder.py
vthorsteinsson: ...and also to be clear: if `pos == end` after the inner loop (which can happen if there is no subtoken match) then you need to increment `pos` so that the outer `while pos< lesc` loop eventually terminates.

tensor2tensor/data_generators/tokenizer.py
vthorsteinsson: Oh, you mean line 60. Agree.

tensor2tensor/data_generators/text_encoder.py
rsepassi: ok, seems reasonable. please add a `tf.logging.warn` (or a log_every_n) when that happens. user needs to know that we don't really support unicode with that encoder. 

we may want to consider moving entirely to bytes in the future and not dealing with unicode at all. but it requires more thought...

tensor2tensor/data_generators/text_encoder.py
rsepassi: ok, thanks

tensor2tensor/data_generators/text_encoder.py
rsepassi: ok, thanks

tensor2tensor/data_generators/text_encoder.py
rsepassi: ok, that's fine

tensor2tensor/data_generators/tokenizer.py
```diff
@@ -45,29 +45,21 @@
...
 class Tokenizer(object):
   """Vocab for breaking words into wordpieces.
   """
 
-  def __init__(self):
-    self._separator_chars = string.punctuation + string.whitespace
-    self._separator_char_mask = array.array(
-        "l", [chr(i) in self._separator_chars for i in xrange(256)])
-    self.token_counts = dict()
+  _SEPARATOR_CHAR_SET = set(string.punctuation + string.whitespace)
```
rsepassi: nice, much cleaner. thank you

tensor2tensor/data_generators/text_encoder.py
rsepassi: Hold on approval; please add this warn.

vthorsteinsson: Will do.

vthorsteinsson: @rsepassi However, there is another possible approach. Python 2 encodes its strings into bytes using UTF-8, so for instance (using some characters from Icelandic, my native language):
```
>>> [ ord(c) for c in "þæðö" ]
[195, 190, 195, 166, 195, 176, 195, 182]
```
In Python3 we can get the same result by explicitly encoding to UTF-8:
```
>>> [ c for c in "þæðö".encode("utf-8") ]
[195, 190, 195, 166, 195, 176, 195, 182]
```
I will modify the pull request to use this approach, which has the advantage of making the Python3 version fully compatible with the Python2 version.

lukaszkaiser:
Having Python 3 give exactly the same results as Python 2 sounds like the best thing, great thanks for this effort Villi!

vthorsteinsson:
I modified ByteTextEncoder to be fully compatible between Python2 and Python3. Both of them encode and decode to an internal UTF-8 representation, although the Python3 one is more explicit about it. This removes the need for special handling of ordinals >= 256.

jgmakin:
I think I see the appeal of mapping any character code outside of 0-255 into an "unknown" character, but that throws out a lot of easily handled cases--which can be done automatically with a package like `unidecode`: it converts en dashes into hyphen-minuses, ellipses into three periods, etc.  That seems preferable for machine-translation data.  Thus, I modified the original `tokenizer.py` very simply with:

```
from unidecode import unidecode
...

token_start = 0

if six.PY2:
    raw_text.decode("utf-8", "replace")
raw_text = unidecode(raw_text)

for pos in range(0, len(raw_text)):
...
```

and then modified `text_encoder.py` only to get rid of the `string-escape` encoding (following this PR).   I haven't tested under python 2.7 but this appears to work under 3.6.

vthorsteinsson:
@jgmakin We ended up with a different solution that uses UTF-8 encoding both on Python2 and 3, with no out-of-bound or "unknown" characters. This seems to work fine on the character generator models. T2T has no difficulty learning correct UTF-8 encoding, apparently.

jgmakin:
Fair enough.  I suppose this issue is shading into questions of problem definition: how many types of characters does a translator have to master?  Maybe someone from the community (with a lot of GPUs) can check how much performance varies under the different conditions: (1) all characters forced into ASCII representation; (2) UTF-8 but with punctuation only from the ASCII set (i.e., the current state of affairs); (3) UTF-8 with properly identified punctuation from the entire UTF-8 set.  My intuition is that (1) should be much easier, but I could be wrong.

vthorsteinsson:
I'm afraid (1) is not really an option if you are translating to or from languages where no ASCII representation exists for important characters; Chinese and other Asian languages are obvious examples and the same applies for Turkish, Icelandic, etc.

jgmakin:
No kidding!  But the popular benchmarks are German-English and French-English.  (I realize that French and German use diacritics, but they are readable without them.  It also seems slightly extreme to me for a translator to have to master distinctions that most literate people are ignorant of, e.g. between hyphens and en dashes.)  Anyway, I concede that using the whole UTF-8 set is the more general solution.

lukaszkaiser:
We run a large number of experiments like that with LSTM models, and it looks like this basically doesn't matter as long as you have at least a bit of various encodings in your training data. So yes -- writing extra code to treat emdash same as dash may give you a little improvement if your training data has very very few emdashes. But it almost doesn't matter in any other setting, and you're putting on yourself the engineering burden and the risk of bugs that come from it. If you want a resilient model, it looked to us to be better to gather some training data with different encodings (or even generate it).

jgmakin:
Alright, I'm convinced! and thanks, "basically doesn't matter..." is interesting.

PR #35, opened 2017-06-24
fix broken reference in inference

jekbradbury:
The class `modality.ClassLabelModality` does not appear to exist (any more?); this change makes inference work for me.


jekbradbury:
Signed CLA.

PR #41, opened 2017-06-24
Remove urllib import (Python3 fix)

vthorsteinsson:
`import urllib`

does not work in Python3. Instead,

`import six.moves.urllib_request`

imports `urllib` on Python2 and `urllib.request` on Python3.


ilya-chistyakov:
Maybe `import six.moves.urllib_request as urllib` will be more correct.

tensor2tensor/data_generators/generator_utils.py
```diff
@@ -22,13 +22,12 @@
...
 import os
 import tarfile
-import urllib
 
 # Dependency imports
 
 import six
 from six.moves import xrange  # pylint: disable=redefined-builtin
-import six.moves.urllib_request
+import six.moves.urllib_request # Imports urllib on Python2, urllib.request on Python3
```
lukaszkaiser: This doesn't seem to work for us with python2 at Google. Could you try this:
import six.moves.urllib_request as urllib

vthorsteinsson:
Fixed with thanks to @ilya-chistyakov. Sorry for the mixup. It works now in my tests both on Python2 and Python3.

lukaszkaiser:
Thanks, that works good.

PR #42, opened 2017-06-25
Use iterator.__next__() for Python3

vthorsteinsson:
This pull request modifies `utils/trainer_utils.py` to use `iterator.__next__()` on Python3 instead of `iterator.next()`, which is only available on Python2.


lukaszkaiser:
Thanks Villi! This looks good but it looks like you might have to get your branch up to date.

vthorsteinsson:
Should be OK now. Sorry for the git mess. You may be able to squash the commits into one on your side - I've had no success with my squashing attempts on my fork.

PR #46, opened 2017-06-25
Use ByteTextEncoding() for encoding strings in character based models

vthorsteinsson:
This PR modifies `wmt.py` to use `text_encoder.ByteTextEncoder()` to **encode** source and target strings in its character based data generators. This synchronizes the behavior with `problem_hparams.py` which uses `text_encoder.ByteTextEncoder()` to **decode** inferred strings. Prior to this PR, encodes and decodes in character models are different and results therefore incorrect.

The PR also defines the constant `EOS` with the value 1 to avoid a "magic constant" appearing in the source code.

### DIFF
--- a/tensor2tensor/data_generators/wmt.py
+++ b/tensor2tensor/data_generators/wmt.py
@@ -33,7 +33,7 @@ import tensorflow as tf
 EOS = 1
 
 
-def character_generator(source_path, target_path, token_vocab=None, eos=None):
+def character_generator(source_path, target_path, token_vocab, eos=None):
   """Generator for sequence-to-sequence tasks that just uses characters.
 
   This generator assumes the files at source_path and target_path have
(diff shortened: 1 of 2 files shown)


vthorsteinsson:
No changes are needed in t2t-datagen since it never calls `character_generator()` directly. Instead, it calls the higher-level functions such as `wmt.parsing_character_generator()` and `wmt.ende_character_generator()`, but the signatures of those haven't changed (and don't need to change).

I will look at eliminating the no-vocab option and update the PR. This will require a change to `wmt_test.py` as well.

vthorsteinsson:
This should be OK now @lukaszkaiser

lukaszkaiser:
Great thanks Villi!

vthorsteinsson:
My pleasure!

PR #57, opened 2017-06-27
algorithmic_reverse_nlplike generator

ReDeiPirati:
@lukaszkaiser I've struggled a bit around Zipf distribution because `numpy.random.zipf` is first of all a Zeta-distribution(are similar but not equal) and second doesn't allow to generate sample from a given range or chose alpha values less than 1.0. So i've followed the advices in this two stackoverflow posts([first](https://stackoverflow.com/questions/1366984/generate-random-numbers-distributed-by-zipf/8788662#8788662) and [second](https://stackoverflow.com/questions/31027739/python-custom-zipf-number-generator-performing-poorly)) and created a function to generate the distribution and another for generating samples(both with test). 
As i said in the closed issue, i have found that alpha(for the Zipf Distr) *usually* is in the range [1.1-1.6] for modelling natural text: so for generating sample which could potentially emulate nlp like task i've chosen(and tested) values that generate samples for all the range following the Zipf distribution.


tensor2tensor/data_generators/algorithmic.py
```diff
@@ -93,6 +93,70 @@ def reverse_generator(nbr_symbols, max_length, nbr_cases):
...
+  Args:
+    distr_map: list of float, Zipf's distribution over nbr_symbols.
+    sample_len: integer, length of sequence to generate.
+
+  Return:
+    sample: list of integer, Zipf's random sample over nbr_symbols.
+
+  """
+  u = np.random.random(sample_len)
+  return [t+1 for t in np.searchsorted(distr_map, u)] # 0 pad and 1 EOS
```
lukaszkaiser: Is 1 enough here for both PAD and EOS? Just asking, if it is, please add a comment making that clear.

tensor2tensor/bin/t2t-datagen
```diff
@@ -86,6 +86,16 @@ _SUPPORTED_PROBLEM_GENERATORS = {
     "algorithmic_multiplication_decimal40": (
         lambda: algorithmic.multiplication_generator(10, 40, 100000),
         lambda: algorithmic.multiplication_generator(10, 400, 10000)),
+    "algorithmic_reverse_nlplike_decimal8K": (
+        lambda: algorithmic.reverse_generator_nlplike(8000, 40, 100000,
+                                                      10, 1.250),
+        lambda: algorithmic.reverse_generator_nlplike(8000, 400, 10000,
```
lukaszkaiser: I think keeping length 40 for both train and dev makes more sense for nlplike tasks. On purely algorithmic tasks, we want to see generalization to much higher lengths. It's a nice-have in NLP too, but less important. Maybe just a little larger, like 60 or so?

tensor2tensor/data_generators/algorithmic.py
ReDeiPirati: It would be enough, but i have continued to think about this line which is a little bit tricky. From the numpy docs about [numpy.random.random](https://docs.scipy.org/doc/numpy/reference/generated/numpy.random.random.html) they said that the values returned are in the range [0.0,1.0) but obtain an absolute zero(0.00000...0) is *possible* but *almost improbable*. So, maybe it's better to add sanity check about the improbable zero. And comment in a more clear way.

tensor2tensor/bin/t2t-datagen
ReDeiPirati: Maybe 70 in train and 700 in dev, would it be better?

PR #59, opened 2017-06-27
Penn TreeBank dataset for language modeling 

pltrdy:
Would like to run PTB benchmark using the `attention_lm`.

I added PTB download & generators. Things about problems_hyperparams and t2t_datagen are not super clear to me, I maybe made mistakes. 

Added a way to work with unknown tokens in `TokenTextEncoder` as well.

### DIFF
--- a/tensor2tensor/data_generators/ptb.py
+++ b/tensor2tensor/data_generators/ptb.py
@@ -1,0 +1,158 @@
+# Copyright 2017 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Data generators for PTB data-sets."""
+
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+import os
+import sys
+import tarfile
+import collections
+
+# Dependency imports
+
+from tensor2tensor.data_generators import generator_utils
+from tensor2tensor.data_generators import text_encoder
+
+import tensorflow as tf
+
+
+EOS = text_encoder.EOS
+PTB_URL = "http://www.fit.vutbr.cz/~imikolov/rnnlm/simple-examples.tgz"
+
+def _read_words(filename):
+  """Reads words from a file.
+      It returns a list of words without '\n'
+      Originally from: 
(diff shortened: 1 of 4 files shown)


tensor2tensor/data_generators/ptb.py
```diff
@@ -0,0 +1,158 @@
...
+  """
+  with tf.gfile.GFile(filename, "r") as f:
+    if sys.version_info[0] >= 3:
+      return f.read().replace("\n", " ").split()
+    else:
+      return f.read().decode("utf-8").replace("\n", " ").split()
+       
+       
+
+def _build_vocab(filename, vocab_path, vocab_size, exclude=[]):
```
lukaszkaiser: Please use `generator_utils.get_or_generate_vocab` instead of this private function. This allows to later run models in multi-problem mode (only possible with matching vocabs).

tensor2tensor/data_generators/text_encoder.py
```diff
@@ -121,15 +121,25 @@ def vocab_size(self):
 class TokenTextEncoder(TextEncoder):
   """Encoder based on a user-supplied vocabulary."""
 
-  def __init__(self, vocab_filename, reverse=False, num_reserved_ids=2):
+  def __init__(self, vocab_filename, reverse=False, num_reserved_ids=2, unk=None):
```
lukaszkaiser: Please use our standard wordpiece-generator instead. It's now a commonly accepted method to get rid of the UNK problem, and it gives better results. On the other hand, if you want to generate vocab just from these files, feel free to change `_DATA_FILES` in `get_or_generate_vocab` or make it a parameter to the function.

pltrdy: That's indeed because of the `_DATA_FILES` constant that I implemented my own `_build_vocab`. I  definitely find more relevant to use a parameter instead. I'll do it.

Thanks for the feedbacks!

jekbradbury:
Unfortunately the PTB language modeling dataset itself, as provided by Mikolov et al., has UNKs for all words outside the top 10k. Even the so-called "character-level PTB" for language modeling has literal sequences of "< U N K >"; all perplexity results reported in the literature on word- or char-level PTB use the dataset with these issues.

pltrdy:
@jekbradbury that's why I did like this. I would like to compare results with RNN-LM as presented in tensorflow's tutorial (that I extended a bit myself). 

I'll probably keep the <unk>, at least on my fork. I don't know where is a >10k vocabulary PTB dataset.

lukaszkaiser:
Indeed, PTB is already pre-processed like this, I forgot, sorry! Let's keep it the way you're doing then and get it, we can make LM1B (the 1 billion word corpus) in a better way, I think that one is detokenized. Thanks and sorry, my mistake with PTB.

pltrdy:
I can still remove the "unk" parameter from TokenTextEncoder as it is useless. One would just need to set the right vocab_size. Handling OOV with a special token isn't really relevant.

pltrdy:
I removed the unk stuff. It is just handled as a regular token. 

I still use my own vocabulary generation. I would make this more generic after I'm sure it is working. I'm now running some experiments.

PR #69, opened 2017-06-29
I'm afraid I need to ask for pull request again for the "Out of index" Error when doing Class_Label_Modality

yynil:
First of all, I apologize for the previous error that I misused the len operator.
Then I figure out if the beam is set to the value that's greater than 1 will still cause the Out of index error.

For Class_Label_Modality search , beam is set to 1 to enable greedy_infere seems to get better results.
@lukaszkaiser 
1. for beam search use the shape.ndims to avoid out of index error
2. for greedy search, still use the shape.ndims to avoid the out of index error. Before that I misuse the slice operation:-(


yynil:
I confirm 1.0.9 T2T class modality decoding works out of box and produces exact the same results I got in this fix. 
Thanks a lot!

PR #72, opened 2017-06-29
Improve error message for missing problems

rshin:
If an incorrect value for `--problems` is specified to `t2t-trainer`, then the list of allowed values printed is sorted in arbitrary order, which is hard to read.


lukaszkaiser:
Thanks for the PR!

PR #75, opened 2017-06-30
Add image augmentation for CIFAR-10

rshin:
* Pad 4 pixels of 0s on each side
* Randomly crop 32x32
* Randomly mirror vertically

Is there a good way to make this optional?


lukaszkaiser:
Thanks for adding this!

PR #100, opened 2017-07-05
Adding attention to LSTM seq2seq baseline

kolloldas:
A new model called **baseline_lstm_seq2seq_attention** which will allow LSTM seq2seq to be used with attention. 

The main component is a cell wrapper called **ExternalAttentionCellWrapper** which is based on the [AttentionCellWrapper](https://www.tensorflow.org/api_docs/python/tf/contrib/rnn/AttentionCellWrapper) but takes attention states externally. The main reason for a cell wrapper is to allow usage with  `dynamic_rnn()`.  It can probably be faster but I wanted to get something out quickly.

I've also created a small hparams set for the attention model.

### DIFF
--- a/tensor2tensor/models/common_layers.py
+++ b/tensor2tensor/models/common_layers.py
@@ -1302,16 +1302,13 @@ def pad_to_same_length(x, y, final_length_divisible_by=1, axis=1):
     return res_x, res_y
 
 
-def pad_with_zeros(logits, labels, return_logits=False):
+def pad_with_zeros(logits, labels):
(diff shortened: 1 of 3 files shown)


lukaszkaiser:
Merged this for now. How are the results? It could be interesting to have a few more parameter settings and try out this model a bit, to know if it works ok.

kolloldas:
Thanks for the merge! I could only run some small tests due some weird problems with the GPU I was working on (It's a Titan X but keeps throwing out of resource errors on all models). Will definitely get some results and tweak the parameters once I solve this GPU issue.

PR #133, opened 2017-07-11
Add new set network layer type and example model

EndingCredits:
Added the following layers (and tests):

*`global_pool_1d`: Not much to this, just a simple max/mean reduce across all vectors in a sequence. Useful for getting a single vector representation of a sequence/set. This can also be thought of as a kind of global attention.
* `linear_set_layer`: Simple 1d convolution across all elements in a sequence (similar layers are already used in repo), but also includes the ability to parameterise the transformation via with a (learned) sequence 'context' vector. Currently this is done by simply concatenating the inputs with the context.
*`ravanbakhsh_set_layer`: Layer type used in https://arxiv.org/abs/1611.04500.

These layers are permutation invariant and hence are suitable for applications where inputs are given as sets, but they may also be useful for sequence tasks as well.

Also added an example model `transformer_alt` which replaces the self attention layers in `transformer` with two different kinds of modules composed of the layers described above. I have no idea how well it performs (although similar previous architectures I tested seemed to do only slightly worse than the full transformer).

N.B: I have not extensively tested these layers in t2t, so it's entirely possible that they're not perfectly functional (particularly wrt to masking).

Also changed two lines in tests to stop numpy warnings.

### DIFF
--- a/tensor2tensor/models/common_layers.py
+++ b/tensor2tensor/models/common_layers.py
@@ -292,7 +292,7 @@ def conv_internal(conv_fn, inputs, filters, kernel_size, **kwargs):
   """Conditional conv_fn making kernel 1d or 2d depending on inputs shape."""
   static_shape = inputs.get_shape()
   if not static_shape or len(static_shape) != 4:
-    raise ValueError("Inputs to conv must have statically known rank 4.")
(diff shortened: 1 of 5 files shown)


lukaszkaiser:
Thanks! Let's get the model clear and training soon.

PR #151, opened 2017-07-13
add wmt_zhen_token_32k

cshanbo:
1. Following the `wmt_ende_token_32k` scheme, where we uses a separated vocabulary (No word segmentation. Please test it, because I found it performs very bad). Both source and target vocabulary contains about 32k words.
2. Using [wmt 2017](http://www.statmt.org/wmt17/translation-task.html) Chinese-English data. Only used `News Commentary v12` as training data, which contains only about 200,000 sentence pairs.
3. In the dev data, there's no plain text data available. We need to process the `sgm` file into plain text first.


lukaszkaiser:
Great thanks @cshanbo ! Are you getting good results?

cshanbo:
Hi @lukaszkaiser 
Sorry that I haven't done enough experiment yet because I don't have the access to enough GPUs. The results now seems not good, I'll let you know if any progress, thank you.

PR #153, opened 2017-07-13
Add Shake-Shake model

rshin:
The model hasn't yet been tested empirically, but it's intended to match "Shake-Shake-Batch" of https://arxiv.org/pdf/1705.07485.pdf on CIFAR-10.

Some potential differences:
* tensor2tensor puts smaller weight decay on larger variables
* Input augmentation details likely differ slightly from fb.resnet.torch
* fb.resnet.torch probably doesn't use gradient clipping
* Different input and output flow. Shake-Shake uses 3x3 conv with 32 channels + batch norm, and 8x8 average pooling + FC, respectively.


lukaszkaiser:
Thanks for the code, we can improve now.

PR #154, opened 2017-07-14
Source/target pair text files; Icelandic parsing support; fixes

vthorsteinsson:
This PR adds text data generators that read source and target sentences in pairs, separated by tabs (\t), from a single source file.

It also adds scaffolding to support parsing of Icelandic text, either from characters or from subword tokens. (More coming later on that front.)

A few smaller fixes are included, such as replacing magic constants to do with reserved token ids with more meaningful identifiers, and getting rid of annoying <pad> strings in decoder output trace messages.


tensor2tensor/data_generators/generator_utils.py
```diff
@@ -290,6 +290,35 @@ def get_or_generate_vocab(tmp_dir, vocab_filename, vocab_size, sources=None):
...
+  """Generate a vocabulary from the source file. This is assumed to be
+  a file of source, target pairs, where each line contains a source string
+  and a target string, separated by a tab ('\t') character. The index
+  parameter specifies 0 for the source or 1 for the target."""
+  vocab_filepath = os.path.join(tmp_dir, vocab_filename)
+  if os.path.exists(vocab_filepath):
+    vocab = text_encoder.SubwordTextEncoder(vocab_filepath)
+    return vocab
+
+  tokenizer = Tokenizer()
```
lukaszkaiser: This line should be removed.

vthorsteinsson:
My pleasure! :-)

lukaszkaiser:
I just found one small problem (source_vocab is used instead of target_vocab in wmt.tabbed_generator). I'll correct it in my next PR and release. It'd be great though to add an option to download the Icelandic parsing set instead of assuming it's there in a file. Is the set public?

Issue #134, opened 2017-07-11
Pluggable datasets


noe:
Currently it is possible to _plug_ you own model for it to be evaluated. This can be done by providing a `--t2t_usr_dir` parameter to `t2t-trainer`, as explained in the [readme](https://github.com/tensorflow/tensor2tensor/#adding-your-own-components), without needing to modify tensor2tensor itself.

However, the same is not possible with datasets. In order to provide your own dataset, you need to modify tensor2tensor itself, as described [data generators README](https://github.com/tensorflow/tensor2tensor/blob/master/tensor2tensor/data_generators/README.md).

It would be nice to have a mechanism for custom datasets that is analogous to that for models.

If it is seen as something good but it's not to be implemented in the near future, are contributions in that direction welcome?

rsepassi:
Thanks @noe. Yes, we're working on making datasets/problems pluggable.

jgevershed:
I believe that the trainer, as well as the datagen, is not 100% "pluggable".
The data generator needs to modify _SUPPORTED_PROBLEM_GENERATORS.
The trainer may also need to add an entry to problem_hparams.PROBLEM_HPARAMS_MAP.

My "solution" is to have my own copies of t2t-trainer and t2t-datagen with minimal changes.
So, I start training by "python my_trainer.py --t2t_usr_dir......." (in Bash on Ubuntu on Windows). 
Luckily the google t2t team has made both these modules very small.
To keep my needed trainer change down to one line, I call a simple helper in my usr dir.

It means I can have my own variant Transformer model and my own problem data, without any hassle in keeping tensor2tensor up to date on my platform.

belph:
@jgevershed Wouldn't something like this work to address this without needing a forked `t2t-datagen`/`t2t-trainer` (if support was added for it, of course)?

```python
@register_problem("image_mnist")
def mnist_problem():
  return (lambda: <generator>, lambda: <generator>)
```

rsepassi:
Currently models, hyperparameter sets, and modalities can be registered without editing the T2T codebase by passing the `--t2t_usr_dir` command line flag. Data generators and problems cannot currently be registered in that way; however, we're working on unifying data generators and problems and adding them to the registry so that it becomes easy to plug in your own dataset & problem without editing the T2T codebase.

rsepassi:
[...]

PR #160, opened 2017-07-17
Support --t2t_usr_dir also in t2t-datagen

noe:
As described in issue #134 , it is now possible to add new problems by subclassing Problem and registering it, but there is no analogous mechanism to parameter `--t2t_usr_dir` from `t2t-trainer` which allows loading user-provided code. This PR simply consists in factoring out the functionality of `--t2t_usr_dir` out of `t2t-trainer` to a common place, and use it from both `t2t-trainer` and `t2t-datagen`.

### DIFF
--- a/tensor2tensor/utils/usr_dir.py
+++ b/tensor2tensor/utils/usr_dir.py
@@ -19,22 +19,12 @@ import sys
 import importlib
 import tensorflow as tf
 
-flags = tf.flags
-FLAGS = flags.FLAGS
 
-flags.DEFINE_string("t2t_usr_dir", "",
-                    "Path to a Python module that will be imported. The "
-                    "__init__.py file should include the necessary imports. "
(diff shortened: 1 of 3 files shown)


tensor2tensor/utils/usr_dir.py
```diff
@@ -0,0 +1,45 @@
...
+# limitations under the License.
+
+"""Utility to load code from an external directory supplied by user."""
+
+import os
+import sys
+import importlib
+import tensorflow as tf
+
+flags = tf.flags
```
rsepassi: could you define the flag in the t2t-trainer as well as the t2t-datagen and have import_usr_dir take a usr_dir as an argument. yes, it'll be duplicated, but generally we try to keep flags in the main "binary" file.

noe: I've updated the PR with the requested changes.

PR #164, opened 2017-07-18
Take in account FLAGS.num_shards

chiphuyen:
To use FLAGS.num_shards in generating data


chiphuyen:
Whoa this is complicated

PR #167, opened 2017-07-19
v1.1.0

rsepassi:


rsepassi:
Lukasz, the failing CLA check is for no-reply@google.com, which is a dummy email. I have an open email thread with the CLA authors to figure out what to do about it but for now I think we ought to ignore it (as we did in a previous PR).

PR #171, opened 2017-07-19
Simplify adding new translation language pairs

martinpopel:
* simplify wmt.py:
`train_generator` is now a method (not a property returning a function)
and WMTProblem subclasses implement it directly (not via one-purpose functions)
* add WMT English-Czech translation problems
* introduce `vocab_name` property, so e.g. `tokens.vocab-en-cs.32768` is used instead of `tokens.vocab.32768` if only English+Czech datasets were used

### DIFF
--- a/tensor2tensor/data_generators/wmt.py
+++ b/tensor2tensor/data_generators/wmt.py
@@ -41,23 +41,6 @@ FLAGS = tf.flags.FLAGS
 # End-of-sentence marker.
 EOS = text_encoder.EOS_ID
 
-
-def _default_token_feature_encoders(data_dir, target_vocab_size):
-  vocab_filename = os.path.join(data_dir, "tokens.vocab.%d" % target_vocab_size)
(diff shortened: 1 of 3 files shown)


tensor2tensor/data_generators/wmt.py
```diff
@@ -69,14 +52,13 @@ def is_character_level(self):
   def targeted_vocab_size(self):
     raise NotImplementedError()  # Not needed if self.is_character_level.
 
-  @property
-  def train_generator(self):
-    """Generator; takes tmp_dir, is_training, possibly targeted_vocab_size."""
+  def train_generator(self, tmp_dir, is_training):
```
lukaszkaiser: Unluckily these arguments now include data_dir, this will need updating. Also, they are different for character models, but it could be just an empty arg (_). Thanks for making this nicer!

martinpopel:
New attempt: #181

PR #176, opened 2017-07-21
Alternative Transformer Fix + Slding Window Attention

EndingCredits:
This includes a fix to stop transformer alternative 'cheating' by attending to elements further on in the sequence. This is equivalent to applying the lower triangular mask (although it does break the permutation invariance of the outputs).

The alternative transformer model performs significantly worse on the en->de translation task, but may have applications to sequential tasks with long sequence lengths as the global pooling is computationally linear in the input sequence length (as opposed to quadratic for self-attention).

I've also added a sliding window attention which basically restricts self-attention to a certain number of elements either side. This could be useful in future for situations with very long input sequences. In my own experiments on a point-cloud dataset, using a window size of 100 for 1000 points (i.e. 1000 element sequence) cuts computation time in half. Unfortunately it's very inefficient for short sequence lengths, and I don't know enough about the internals of tensorflow to improve it.

EDIT: I notice this is similar to the block attention which was added recently, will need to take a closer look at this.

This is from quite an old branch, so I can look at updating to latest commit if needed.


EndingCredits:
As I mentioned, I tried to add a windowed attention with the purpose of reducing computational complexity. I think the output of both methods are basically the same, but it's possible that my method is slightly more efficient (since I designed it with that in mind). That said, the way I do it is less optimised to the hardware, and the only difference is having to do an n x m calculation rather than an n x n followed by masking (it depends how much of the work is the backprop). So I'll have to look into the performance profiling.

I can definitely look into combining both methods, so I'll try and merge them into the same function.

Out of interest, what was your motivation for windowed attention?

lukaszkaiser:
Our main motivation was to fit into GPU memory when trying to run on long sequences (like the `wiki_32k` problem). But less computation is also good to have. Take a look and ping me please when you think it's ready to merge!

EndingCredits:
@lukaszkaiser I've updated the local masked attention to do what I wanted to do with sliding window attention. Basically it restricts attention to elements within a certain window, which is either n elements left, or n/2 elements left and right (rather than a whole number of blocks). I believe the older functionality (and also unmasked) can be restored simply by changing the mask, but I'm not 100% sure since I wasn't able to work out exactly how they worked.

EndingCredits:
Okay, merged all the different types into a single function, the one thing that isn't finished is getting the bias add working properly, but the vanilla mask right works fine.  Should be ready for merge.

lukaszkaiser:
Hi William. I reverted some parts of your change in 1.1.4. Mostly naming stuff: kept `masked_local_attention` and `unmasked_local_attention` because we still have some code running around that needs them. But the unmasked one is the same as you proposed, just without the (not-working) bias for now. I'll test it a bit more and remove the unmasked version in time. Thanks for looking at it, and keep in touch with the alternative transformer results. Thanks!

EndingCredits:
From what I understand, `masked_local_attention` should also be the same as `local_attention` with `look_right` set to False and `use_whole_block` set to true. With those settings the code should be pretty much identical except I use padding to get around the first block problem. I'm a lot more uncertain on the similarity with `unmasked_local_attention`. But probably sensible to keep them separate until we know for sure anyway.

PR #182, opened 2017-07-26
use s.endswith() where appropriate

martinpopel:
Most translation language pairs involve English
and both the source and target filenames contain substring "en"
(e.g. europarl-v7.de-en.de), so
`if "en" in filepath` was always evaluated as True.


lukaszkaiser:
I hand-merged this into 1.1.3. Could you verify that all parts are there? Thanks again for correcting this!

martinpopel:
Yes, thanks.

lukaszkaiser:
Thank you!

PR #185, opened 2017-07-27
v1.1.2

rsepassi:


rsepassi:
The CLA errors are again all for the T2T Team author, which is a dummy email. Ok to merge.

PR #206, opened 2017-08-03
Algorithmic Shift and Vigenere Cipher Data Generator

aidangomez:
Adds a simple shift cipher and vigenere cipher on synthetically sampled data.


aidangomez:
Should be fixed!

PR #248, opened 2017-08-23
Bigger translate_encs_wmt32k training data, tsv support

martinpopel:


tensor2tensor/data_generators/generator_utils.py
```diff
@@ -333,19 +333,18 @@ def generate():
...
-      with tarfile.open(compressed_file, read_type) as corpus_tar:
-        corpus_tar.extractall(tmp_dir)
-
       for lang_file in source[1]:
         tf.logging.info("Reading file: %s" % lang_file)
         filepath = os.path.join(tmp_dir, lang_file)
+        if not tf.gfile.Exists(filepath):
+          read_type = "r:gz" if filename.endswith("tgz") else "r"
+          with tarfile.open(compressed_file, read_type) as corpus_tar:
+            corpus_tar.extractall(tmp_dir)
```
rsepassi: why did you move this down here into the loop? shouldn't it conceptually happen before the loop runs?

tensor2tensor/data_generators/wmt.py
```diff
@@ -267,6 +269,10 @@ def bi_vocabs_token_generator(source_path,
 # English-Czech datasets
 _ENCS_TRAIN_DATASETS = [
     [
```
rsepassi: this doesn't seem to be a WMT file. maybe this makes sense as a new problem instead of adding onto wmt.

tensor2tensor/data_generators/generator_utils.py
martinpopel: This is intentional - it prevents extracting the tar, if it has been already extracted, so it saves time (if multiple Problems share the same tar).
More importantly, it is necessary if the input files are not available online, but I copied them to the `tmp` directory (and didn't want to pack them into tar, but I still want T2T to create the vocabulary).

tensor2tensor/data_generators/wmt.py
martinpopel: The CzEng corpus is officially part of the WMT training data for English-Czech, see http://statmt.org/wmt17/translation-task.html.
Without CzEng you cannot get the SotA WMT scores (it is much bigger than news-commentary-v12.cs-en and europarl.cs-en).

tensor2tensor/data_generators/generator_utils.py
martinpopel: And now, reading my 6e825430c776 commit message, I remember a third reason: the files in the CzEng tar don't have write permission (-r--r--r--). So if you try to extract this tar file and it is already extracted, it will fail (and the whole t2t-datagen as well).

PR #250, opened 2017-08-24
reserve also UNK token, more robust TokenTextEncoder

martinpopel:
TokenTextEncoder is used in data_generators.wmt.ende_bpe_token_generator()
in the translate_ende_wmt_bpe32k problem, which uses an external BPE vocabulary
and training data pre-processed with these BPE subwords.
This may be useful when comparing with Nematus/NeuralMonkey etc. on exactly the same training data.
Unfortunately, the TokenTextEncoder implementation was not robust enough.
It failed if there was an unknown subword in the training or dev or test data,
e.g. because of unknown character.

This PR solves the main problem (failure) for me, but
- UNK tokens are silently ignored without any warning
- NUM_RESERVED_TOKENS is increased, so it may break compatibility with legacy models
  (What about setting NUM_RESERVED_TOKENS high enough for any future use?)

Feel free to close this PR without merging if you don't like it.

### DIFF
--- a/tensor2tensor/data_generators/text_encoder.py
+++ b/tensor2tensor/data_generators/text_encoder.py
@@ -38,10 +38,12 @@ import tensorflow as tf
 # Reserved tokens for things like padding and EOS symbols.
 PAD = "<pad>"
 EOS = "<EOS>"
-RESERVED_TOKENS = [PAD, EOS]
+UNK = "<UNK>"
+RESERVED_TOKENS = [PAD, EOS, UNK]
 NUM_RESERVED_TOKENS = len(RESERVED_TOKENS)
 PAD_ID = RESERVED_TOKENS.index(PAD)  # Normally 0
 EOS_ID = RESERVED_TOKENS.index(EOS)  # Normally 1
+UNK_ID = RESERVED_TOKENS.index(UNK)  # Normally 2
 
 if six.PY2:
   RESERVED_TOKENS_BYTES = RESERVED_TOKENS
@@ -171,7 +173,7 @@ class TokenTextEncoder(TextEncoder):
 
   def encode(self, sentence):
     """Converts a space-separated string of tokens to a list of ids."""
-    ret = [self._token_to_id[tok] for tok in sentence.strip().split()]
+    ret = [self._token_to_id.get(tok, UNK_ID) for tok in sentence.strip().split()]
     return ret[::-1] if self._reverse else ret
 
   def decode(self, ids):


lukaszkaiser:
Hi Martin. We talked a bit about this, and I think we'd prefer to avoid adding UNK as the default. There are 2 main reasons. For one, a fair bit of datasets work perfectly fine without UNK: all algorithmic/toy data, every problem that uses subwords, all image problems, genetics, and so on. On the other hand, it's quite easy to just add UNK to your own vocabulary. For example, you can set the class for the BPE problem to have 3 reserved tokens and do UNK, without changing all other problems. In general, we're moving away from BPE and I don't think we should force all problems to have an UNK just because of it (note that our SubwordTextEncoder will correctly encode any string into an allowed vocabulary, even if it has letters outside the training set, using byte-escape sequences). So maybe make this PR smaller, e.g., just apply to the BPE problem? We also need to port BPE to the Problem class, but that's a parallel point (it'll make it easier to change that in one place, for now it needs to be changed in wmt.py and in problem_hparams). What do you think?

martinpopel:
@lukaszkaiser Thanks for your comments.

> For one, a fair bit of datasets work perfectly fine without UNK:

I know, but I thought that an extra reserved ID makes no harm, even if it is unused in most models. I admit, it may be confusing for the users.

> note that our SubwordTextEncoder will correctly encode any string into an allowed vocabulary, even if it has letters outside the training set, using byte-escape sequences

I know, it's great. I wanted to compare this with an external BPE vocabulary (in order to make this experiment as comparable to Nematus as possible).

> t's quite easy to just add UNK to your own vocabulary.

Yes, adding one line to the vocabulary file is easy, but I need to change the code, so it returns the ID of this line if the token is unknown.
I know I could preprocess all the training, dev and test data files and convert all out-of-vocabulary tokens to UNK,
but I felt it is easier and more elegant to do this within T2T code.

> So maybe make this PR smaller, e.g., just apply to the BPE problem

Do you mean to subclass TokenTextEncoder as e.g. TokenTextEncoderWithUnk?
I can do this, but I still feel it would be better to keep NUM_RESERVED_TOKENS the same for all problems
(it is defined globally within text_encoder.py, not as a property of TokenTextEncoder instances).

> We also need to port BPE to the Problem class, but that's a parallel point (it'll make it easier to change that in one place, for now it needs to be changed in wmt.py and in problem_hparams). What do you think?

Yes, I planned to send another PR with this (converting `translate_ende_wmt_bpe32k` to the Problem class).

Anyway, I am closing this PR. I need it only for one experiment, I don't need it merged in master.

anglil:
I preprocessed the training and test data and replace oov words with an UNK sign. After doing this data-gen and training both worked.

lukaszkaiser:
I added an option to `replace_oov` in 1.2.1 and moved the BPE version to Problem and added UNK to vocabulary. Could you take a look and check if that works for you? Thanks!

PR #261, opened 2017-08-29
Add unknown token support to TokenTextEncoder

rmbrad:
I needed to support out-of-vocabulary terms for some NER experiments but like mentioned in #250 the TokenTestEncode would fail on OOV terms. This PR updates TokenTestEncoder to allow problems to specify a token to be returned for OOV terms. 

It is similar to #250 except it allows problems to opt-in by specifying the `unknown_token` parameter and it doesn't add a new reserved token. Since a reserved token is not added it is the problems responsibility to ensure the unknown token exists in the vocabulary.

### DIFF
--- a/tensor2tensor/data_generators/text_encoder.py
+++ b/tensor2tensor/data_generators/text_encoder.py
@@ -73,6 +73,35 @@ else:  # No conversion required on Python >= 3.
     return s
 
 
+class unkdict(dict):
+  """Returns specified value for unknown keys, but does not add to dictionary.
+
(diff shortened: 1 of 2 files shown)


lukaszkaiser:
I added an option to `replace_oov` in 1.2.1 and corrected `translate_ende_bpe32k` to use it. Could you give it a look and check if that's enough for your use-case? Thanks!

rmbrad:
@lukaszkaiser `replace_oov` covers my use case with no problem, I'll go ahead and close this PR. Thanks!

PR #289, opened 2017-09-08
tensor2tensor: xrange(10) --> range(10) for Python 3

cclauss:
xrange() was removed in Python 3.  For counting to 10, `from six.moves import xrange` would be overkill.


rsepassi:
Thanks!

PR #304, opened 2017-09-13
Fix key deletion during key iteration in Python 3

soaxelbrooke:
Deletion of dict keys while iterating over said dict is a no-no in Python 3, and RuntimeErrors.  This simply consumes the dict key iterator into a list before iterative deletion.


vthorsteinsson:
Alternative, even more concise:

`d = { k : v for k, v in d.items() if v is not None }`

rsepassi:
Thanks!

PR #362, opened 2017-10-15
added .vscode to gitignore and correct the decoded outputs and targets, and a func typo

cuiyong0915:
Now the console output is:
Inference results TARGET: a b  c d  ...

It should be:
Inference results TARGET: ab cd ...


cuiyong0915:
I signed it!

cuiyong0915:
It has been fixed in master.

PR #370, opened 2017-10-19
Rework the translate problem

vince62s:
Rename wmt to translate, make it a single class file.
Split language pairs for clarity in each translate_enxx.py
Dissociate ende / enfr, make them independant, no real reason to combine them.

For enfr, I commented the huge WMT dataset and replaced it with a baseline 1M segment we use in OpenNMT. Good for benchmarking too :)


vince62s:
I signed it!

vince62s:
@lukaszkaiser please review.

tensor2tensor/data_generators/generator_utils.py
```diff
@@ -338,13 +338,19 @@ def generate():
 
         # Use Tokenizer to count the word occurrences.
         with tf.gfile.GFile(filepath, mode="r") as source_file:
-          file_byte_budget = 3.5e5 if filepath.endswith("en") else 7e5
+          file_byte_budget = 1e6 if filepath.endswith("en") else 1e6
+          counter = 0
+          countermax = int(source_file.size() / 1e6)
```
martinpopel: I understand the goal: each `countermax`th line is sampled, but
* I would suggest to make `countermax` lower, otherwise in 50% of cases we reach the end of file without reaching file_byte_budget bytes of sampled data (I expect line lengths are distributed randomly in the training data).
* Don't repeat the constant 1e6, use `file_byte_budget` instead.
* Add a comment about the intended goal: to get a representative sample from (possibly) unshuffled training data

tensor2tensor/data_generators/generator_utils.py
```diff
@@ -338,13 +338,19 @@ def generate():
...
-            yield line
+            if counter < countermax:
+              counter += 1
+            else:
+              if file_byte_budget <= 0:
+                break
+              line = line.strip()
+              file_byte_budget -= len(line)
+              counter = 0
+              yield line
```
martinpopel: * Maybe it would be better to have this file_byte_budget as a separate PR as it is not really related to wmt.py refactoring and (unlike the refactoring) it changes the behavior and may affect BLEU.
* Now I realized that even better than solution would be to increase file_byte_budget dynamically as long as the subword/BPE algorithm ends up with min_count=1 (or 2 or a user-specified constant). Or (to make it faster) we could make sure the space-delimited-tokens (or e.g. 10-bytes-tokens for non-space languages) vocabulary size is significantly bigger than  the expected subword vocab size.

lukaszkaiser: Let's not make it too complicated. Using a parameter `file_budget` (per file) should be good enough for great most cases, no?

lukaszkaiser:
I think it looks good to merge, but would like to wait until Martin's comments are addressed, let me know!

vince62s:
ok guys, let make this simple, we can still adjust later on.
I adjusted countermax to take into account Martin's comment on the file distribution.
As is now, it stays similar to the original concept, which is the same file_byte_budget for each file.
Let's get some results of a few training, but my feeling is that is good enough for most cases.

mehmedes:
Just a thought: What do you think of applying bi_vocabs not only for enzh but also for all other problems that are bilingual?

colmantse:
according to my experiment on web crawl, straight forward subword works fine, there is no need of bivocab given sufficient datasize.

martinpopel:
@mehmedes: I don't think bi_vocab is a good idea because it prevents segmenting named entities the same way in both languages (and thus translating them correctly/acceptably). Even for English-Chinese it is an open question whether bi_vocab gives better results (probably depending on data).

lukaszkaiser:
I think we should make it easy to use bi-vocab, but keep it all on a problem-to-problem basis. There might also be more work on the vocab side and to make the budget easier to tweak. But I think this PR is large enough and good as it it. I think it's ready to merge, should I wait for anything more? I'll wait an hour or so and merge then if noone complains. Great thanks for doing this!

vince62s:
yes it's ok to merge.

lukaszkaiser:
Thanks guys!

Issue #375, opened 2017-10-24
Double linear projection for attention states in lstm


lucasgreene:
I've been working with the lstm modules and was wondering about the attention cell wrapper in models/lstm.py

```
 def call(self, inputs, state):
    """Long short-term memory cell with attention (LSTMA)."""

    if not isinstance(state, AttentionTuple):
      raise TypeError("State must be of type AttentionTuple")

    state, attns = state
    attn_states = self._attn_states
    attn_length = attn_states.get_shape()[1].value
    if attn_length is None:
      attn_length = tf.shape(attn_states)[1]

    input_size = self._input_size
    if input_size is None:
      input_size = inputs.get_shape().as_list()[1]
    if attns is not None:
      inputs = tf.layers.dense(tf.concat([inputs, attns], axis=1), input_size)
    lstm_output, new_state = self._cell(inputs, state)

    new_state_cat = tf.concat(nest.flatten(new_state), 1)
    new_attns = self._attention(new_state_cat, attn_states, attn_length)

    with tf.variable_scope("attn_output_projection"):
      output = tf.layers.dense(
          tf.concat([lstm_output, new_attns], axis=1), self._attn_size)

    new_state = AttentionTuple(new_state, new_attns)

    return output, new_state
```

So the basic logic here is take the current computed attention vector (computed during the previous call to call()), linearly combine it with the current inputs, compute a new lstm state and output, create the next attention state from the new lstm state. All good until here:
```    
with tf.variable_scope("attn_output_projection"):
      output = tf.layers.dense(
          tf.concat([lstm_output, new_attns], axis=1), self._attn_size)
```
Here we are linearly projecting the new attention state with the lstm output. Is there justification for this in a paper somewhere? Im just a little confused because if you are using this function as a generating decoder and feeding in the previous lstm output as the next input you will be combining the output/input with the new attention state twice. Im wondering because I am training a model that takes english sentences as both input and output and I am finding the model tends to repeat itself a good bit when the input and outputs are very similar. The thought here is maybe the attention mech being applied twice is causing it to keep copying the same word from the input source. Any ideas welcome!

kolloldas:
The original [Bahdanau](https://arxiv.org/pdf/1409.0473.pdf) paper only had input attention. But later papers have found output attention helps the model as well. Please check [Luong](https://arxiv.org/pdf/1508.04025.pdf). Essentially:
* Combine the attention (context vector) from the previous time step with the current inputs
[...]

Issue #205, opened 2017-08-03
Slow lstm_seq2seq_attention training


fstahlberg:
I am trying to run the attention-based seq2seq baseline on wmt_ende_tokens_32k by just using

  MODEL=lstm_seq2seq_attention
  HPARAMS=lstm_attention

in the walkthrough. However, training seems to be very slow: the global step/sec rate drops down to around 0.1 on a single K80, and the GPU utility is around 15%. Using a much smaller batch size by adding --hparams='batch_size=32'  to the t2t-trainer command helps a bit but the GPU utility is still around 25%. Am I missing something? Thanks.

colmantse:
I have similar issue, i am sure that my gpu is running but my cpu is also operating on 98%, i wonder if it is normal and it is taking ages as compared with transformer.

2877992943:
yes ,same to me, compared with transformer with 22 seconds for 100 steps, it took 600 seconds 
 every 100 steps for lstm_attention

kolloldas:
Can you check the latest version? It now uses the `AttentionWrapper` from *seq2seq* library which is much more optimized than the previous implementation. 
For the `translate_ende_wmt_bpe32k` task I'm getting ~5steps/s on a K80

rsepassi:
Thanks @kolloldas!
Transformer will generally be faster.
And hopefully the switch to the seq2seq libs will speed up the LSTM.
Closing for now.

PR #377, opened 2017-10-25
Update LSTM Attention Model to use tf.contrib.seq2seq.AttentionWrapper

kolloldas:
The LSTM Attention model was updated:
1. Replaced old attention model with `AttentionWrapper` based code
2. *HParams* added for Bahdanau Type Attention and Luong Type Attention
3. *HParams* added for multi-head attention
4. Set weight decay to 0 to fix NaN training loss issues

I did a quick comparison with the variants on the *translate_ende_wmt_bpe32k* task. 
After 2000 train steps:

| Model                                 | Parameters | Validation Loss  | BLEU | Accuracy |
| ----------------------------- | ------------- | -----------------|-------|-----------|
| LSTM Bahdanau Attention  | 14867456 | 6.25 | 0.2351 | 11.73% |
| LSTM Luong Attention        | 14850944  | 6.27 | 0.2319 | 11.72% |
| LSTM Bahdanau Attention Multi (4)  | 15228416 | 6.18 | 0.2370 | 11.99% |
| LSTM Luong Attention Multi (4)       | 15211520 | 6.16 | 0.2358 | 12.27% |
| Transformer (Base Single GPU)        | 63103488 | 5.95 | 0.1223 | 14.22% |

Please review, thanks!

### DIFF
--- a/tensor2tensor/models/lstm.py
+++ b/tensor2tensor/models/lstm.py
@@ -31,144 +31,6 @@ from tensor2tensor.utils import t2t_model
 import tensorflow as tf
 from tensorflow.python.util import nest
 
-# Track Tuple of state and attention values
-AttentionTuple = collections.namedtuple("AttentionTuple", ("state",
-                                                           "attention"))
-
-
-class ExternalAttentionCellWrapper(tf.contrib.rnn.RNNCell):
-  """Wrapper for external attention states for an encoder-decoder setup."""
-
-  def __init__(self,
-               cell,
-               attn_states,
-               attn_vec_size=None,
-               input_size=None,
-               state_is_tuple=True,
-               reuse=None):
-    """Create a cell with attention.
-
-    Args:
-      cell: an RNNCell, an attention is added to it.
-      attn_states: External attention states typically the encoder output in the
-        form [batch_size, time steps, hidden size]
-      attn_vec_size: integer, the number of convolutional features calculated
-        on attention state and a size of the hidden layer built from
-        base cell state. Equal attn_size to by default.
-      input_size: integer, the size of a hidden linear layer,
-        built from inputs and attention. Derived from the input tensor
-        by default.
-      state_is_tuple: If True, accepted and returned states are n-tuples, where
-        `n = len(cells)`.  Must be set to True else will raise an exception
-        concatenated along the column axis.
-      reuse: (optional) Python boolean describing whether to reuse variables
-        in an existing scope.  If not `True`, and the existing scope already has
-        the given variables, an error is raised.
-    Raises:
-      TypeError: if cell is not an RNNCell.
-      ValueError: if the flag `state_is_tuple` is `False` or if shape of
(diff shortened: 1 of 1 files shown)


martinpopel:
Is there any explanation for the disagreement between BLEU and Accuracy (or Validation loss)? Transformer has the best accuracy, but the worst BLEU according to your table.

kolloldas:
@martinpopel there appears to be some problem with the way BLEU is computed by t2t, almost as if lower is better! Here are the results after 5000 train steps. I've now included Rouge-L scores as well:

| Model                                 | Parameters | Validation Loss  | BLEU | Accuracy | Rouge-L |
| ----------------------------- | ------------- | -----------------|-------|-----------|---------|
| LSTM Bahdanau Attention  | 14867456 | 5.76 | 0.1818 | 14.58% | 0.169 |
| LSTM Luong Attention        | 14850944  | 5.75 | 0.1965 | 14.52% | 0.168 |
| LSTM Bahdanau Attention Multi (4)  | 15228416 | 5.67 | 0.1394 | 14.84% | 0.175 |
| LSTM Luong Attention Multi (4)       | 15211520 | 5.64 | 0.1311 | 15.47% | 0.179 |
| Transformer (Base Single GPU)        | 63103488 | 4.73 | 0.0636 | 24.02% | 0.279 |

As you can see the differences between transformer and the rest are more pronounced.

vince62s:
I think you need to wait 50k or 100k to get meaningful values.
But a few questions:
What learning rate / decay did you take for LSTM ?
Do I read well saying you took 2x128 as default for layers / hidden size ?

kolloldas:
Yes that's true, 5K steps are not enough, but it was more of a sanity check.
I kept the default hyperparameters in T2T:
learning_rate = 0.1
num_layers = 2
hidden_size = 128

kolloldas:
Great! 
Should be able to close #113 #205 #375 #378

martinpopel:
@kolloldas Do the approx_bleu values after 100k steps look reasonable?

I've noticed the approx_bleu curve always starts from quite high values at 0 steps (sometimes even 0.8), then drops (e.g. to 0.05 at 10k steps) and then starts growing and only from that point it correlates with the real BLEU.
Does anyone know why?
Is it possible that approx_bleu is computed in a way that when predicting the n-th (subword) token, tokens 1..n-1 *from the reference translation* are known (and backpropagated), so the network first tries to repeat the last token (and shifting words by one has just a minor effect on BLEU).

kolloldas:
I'm yet to run this task for 100K steps but I think BLEU is initially high because the model makes very short predictions at the beginning. The BLEU drops as predictions get longer (but are mostly gibberish) and then finally begins to pick up as predictions become more meaningful..

martinpopel:
No. BLEU contains a *brevity penalty* term, so too short translations get very low BLEU. I think the reason is what I have described above, ie. approx_bleu by default is not computed autoregressively, but using previous gold words as input for the decoder. You can use `--eval_run_autoregressive`, but this seems to be currently buggy (#407).

Issue #353, opened 2017-10-12
Restoring models hangs (sometimes)


epurdyf:
I have a model of moderate complexity written in tensorflow using tensor2tensor. (It uses an RNN, if that is relevant. All other parts are pretty trivial.) The model saves to disk relatively quickly (~30-60 seconds), but when I try to restore it, it hangs, seemingly forever. I've given it half an hour and it never finishes.

Interestingly, it is able to restore a model relatively efficiently when the job first starts. When I examine the running processes using top, the successful loading takes up 300% cpu, while the unsuccessful loading takes up about 30% cpu, and that only sporadically.

I have tried to make sure that I am using the C++ implementation of protobuf, as that is one possible cause of slow loads. Not entirely sure how to verify that I am using it, however.

epurdyf:
Bizarrely, enough, the problem seems to go away if I avoid the function tf.nn.bidirectional_dynamic_rnn and roll my own bidirectional RNN.

lukaszkaiser:
That's very bizzare and looks more like a generic TF issue than a T2T one. Can you reproduce the problem with a pure TF test? In that case, it'd be great to submit a bug to Tensorflow. Otherwise, we'd need you to make a PR with your model to reproduce. Thanks for reporting in all cases!

epurdyf:
Finally managed to get a small, clean reproduction of the bug. It seems that any/most models involving an LSTM are doing something really weird when using the default "orthogonal" initializer.

https://github.com/tensorflow/tensor2tensor/pull/381

PR #381, opened 2017-10-26
Minimal example reproducing bug in issue #353

epurdyf:


lukaszkaiser:
For now I'm adding a check that for LSTMs the initializer is not orthogonal. It's strange to me, but I think this is a generic TF issue with orthogonal initializer. It'd be great to make a minimal pure-TF (non-T2T) example and report a bug in TF.

PR #398, opened 2017-11-03
Introduce Max Decode Length

resec:
Added the `max_decode_length` decoding parameter to . It defaults to `0`, which is disabled, if set > 1, then will restrict the max decode steps with respect to both input length and `extra_decode_length`

And the original `decode_length` is changed to `extra_decode_length` to avoid confusion.

This would be useful in case where we have a very long "inputs" but only expecting short "outputs" (e.g. input length in length [10, 1000], but only wants <=20 output length)

One more thing is in the interactive decoding, nothing is changed with respect to the `max_decode_length`, please let me know if should.


lukaszkaiser:
I'm not sure I understand this. Why not have `decode_length` as now in the class and just add the input length to it at the start? Having 2 lengths (`decode_length` and `extra_decode_length`) seems like an overkill, or am I missing something?

lukaszkaiser:
I'm not sure I understand this. Why not have `decode_length` as now in the class and just add the input length to it at the start? Having 2 lengths (`decode_length` and `extra_decode_length`) seems like an overkill, or am I missing something?

lukaszkaiser:
I'm not sure I understand this. Why not have `decode_length` as now in the class and just add the input length to it at the start? Having 2 lengths (`decode_length` and `extra_decode_length`) seems like an overkill, or am I missing something?

resec:
@lukaszkaiser Let me try to explain more to you.

Previously the `decode_length` actually have two meanings in the code, one is "extra decode length" that is used to add upon input length for decoding. While in some other code, it can also be the institutive "decode length" (e.g. beam_search's decode length). So first of all this is very confusing.

As the interface provided in `decoding.decode_hparams` only have "extra decode length", which is by default adding it to input length as the actual decode length, so I rename all `decode_length` where is actually meaning "extra decode length" to clear the confusion. (What I did in the first commit)

But actually, above is the first step for my true intention.

As mentioned in gitter. In cases where we want to decode a short output that is much smaller input length, without the `max_decode_length`, we are not able to do it as far as I know. So I added `max_decode_length`.

Hope I am explaining well. Thanks.

lukaszkaiser:
Sorry, but I still don't understand the difference. You say that: one is "extra decode length" that is used to add upon input length for decoding, while in some other code, it can also be "decode length", e.g. beam_search's decode length. I think these 2 things are the same, it's a global bound on the maximum length of the decoded sequence, i.e., it's the same as `max_decode_length` that you're adding.

Why not, what am I missing? If it is the same, then I'd be happy to rename it and not add to input length (which is confusing), but have just 1 parameter, `max_decode_length`. Why not, what's the problem?

resec:
@lukaszkaiser "adding the extra decode length to input length" actually is not use case to me. For my case, I do just want a way to control the final "decode length". 

But not sure if other users are relying on the "adding the extra decode length to input length". Maybe in some cases, users want to decode some more symbols upon input length. But once again, I am not sure.

I am fine to remove the "extra decode length" feature, and willing to implement the change if needed.

Thanks a lot.

lukaszkaiser:
I think this got obsolete. Let's close this one and open a new, simpler one if still needed.

Issue #349, opened 2017-10-10
How to decode against WMT ENDE model via Tensorflow Serving


robertBrnnn:
Somewhat similar to https://github.com/tensorflow/tensor2tensor/issues/344, I have my model served by TensorFlow Serving API, how can I translate an English sentence against this model now?

Model created by setting `export_saved_model=True` during training, and served by `tensorflow_model_server --model_base_path=/the/path/ --port=9000`.

robertBrnnn:
So I've gone further into the abyss, trying to figure this out, and have gotten to the point of sending a predict request to my served model.

Took quite a bit of digging through the t2t source code but I've encoded the text as `input_ids = vocabulary.encode(text)` and converted to array by `np.array(x).astype(np.int32)`. 

I construct request as `request.inputs['inputs'].CopyFrom(
        tf.contrib.util.make_tensor_proto(input_ids, dtype=tf.int32, shape=????))` but don't know what to put in for shape, from going through t2t's code I've found reshape calls of either `[1, 299, 299, -1]` or `[1, -1, 1, 1]`, neither of these work and the error I receive is `ValueError: Too many elements provided. Needed at most -89401, but received 4` and `ValueError: Too many elements provided. Needed at most 1, but received 4` respectively. I've guessed that the max elements it requires is determined by the multiplication of elements 1 and 2 in the arrays. If I increase these, the request just hangs. 
By leaving shape unspecified the error is `AbortionError(code=StatusCode.INVALID_ARGUMENT, details="Expected multiples argument to be a vector of length 2 but got length 5` 

Any help would be much appreciated.
R.

ninotoshi:
What is the result of `saved_model_cli show`?

In my case,

```
  request.inputs['inputs'].CopyFrom(tf.contrib.util.make_tensor_proto([[[[1]]], [[[2]]]]))
```
worked for this model
```
# saved_model_cli show --dir /path/to/saved/model/dir --tag_set serve --signature_def serving_default
The given SavedModel SignatureDef contains the following input(s):
inputs['inputs'] tensor_info:
    dtype: DT_INT32
    shape: (-1, 1, 1, 1)
    name: ExpandDims_2:0
The given SavedModel SignatureDef contains the following output(s):
outputs['outputs'] tensor_info:
    dtype: DT_INT32
    shape: (-1, -1)
    name: strided_slice_8:0
Method name is: tensorflow/serving/predict
```

The response is
```
(outputs {
  key: "outputs"
[...]

PR #400, opened 2017-11-03
2-Dims placeholder for tf.VarLenFeature

resec:
Currently, the outputted `SavedModel` as decode server, the generated placeholder shape for tf.VarLenFeature will be a `[None]` int array. Then by intuition if we feed with a 1-Dim int array, the input will be transformed to a `[None, 1, 1, 1]` via `tf.expand_dims`. Then the only one int array input is treated as a batched int input, and everything else outputted form model will be crashed.

By adding one rank with undefined dim size to placeholder will fix this.

I tired `transformer`, `transformer_vae`, and `lstm`, all hit this problem, but not sure if all other models are the same. I am not fermilier with all other models and possiblly this change would break other models.


resec:
I think #349 should be related.

Issue #397, opened 2017-11-03
fast decoding error when running t2t_decoder


fhieber:
I am trying to decode a custom problem using standard transformer hyper-parameters.
Since 1.2.6 I am receiving the following error when running t2t-decoder:
```
...
INFO:tensorflow:decode_hp.batch_size not specified; default=32
INFO:tensorflow:Performing decoding from a file.
INFO:tensorflow:Getting sorted inputs
INFO:tensorflow: batch 94
INFO:tensorflow:Decoding batch 0
WARNING:tensorflow:Input graph does not use tf.data.Dataset or contain a QueueRunner. That means predict yields forever. This is probably a mistake.
WARNING:tensorflow:Schedule=train_and_evaluate. Assuming that training is running on a single machine.
INFO:tensorflow:datashard_devices: ['gpu:0']
INFO:tensorflow:caching_devices: None
INFO:tensorflow:Beam Decoding with beam size 5
Traceback (most recent call last):
  File "/usr/local/bin/t2t-decoder", line 103, in <module>
    tf.app.run()
  File "/usr/lib/python3.4/dist-packages/tensorflow/python/platform/app.py", line 48, in run
    _sys.exit(main(_sys.argv[:1] + flags_passthrough))
  File "/usr/local/bin/t2t-decoder", line 92, in main
    FLAGS.decode_to_file)
  File "/usr/local/lib/python3.4/site-packages/tensor2tensor/utils/decoding.py", line 235, in decode_from_file
    for result in result_iter:
  File "/usr/lib/python3.4/dist-packages/tensorflow/python/estimator/estimator.py", line 412, in predict
    features, None, model_fn_lib.ModeKeys.PREDICT, self.config)
  File "/usr/lib/python3.4/dist-packages/tensorflow/python/estimator/estimator.py", line 698, in _call_model_fn
    model_fn_results = self._model_fn(features=features, **kwargs)
  File "/usr/local/lib/python3.4/site-packages/tensor2tensor/utils/model_builder.py", line 349, in wrapping_model_fn
    return model_fn(model, features, mode, hparams, **kwargs)
  File "/usr/local/lib/python3.4/site-packages/tensor2tensor/utils/model_builder.py", line 164, in model_fn
    max_idx=len(hparams.problems) - 1)
  File "/usr/local/lib/python3.4/site-packages/tensor2tensor/utils/input_fn_builder.py", line 182, in cond_on_index
    return fn(cur_idx)
[...]

PR #404, opened 2017-11-06
add use_last_position_only=True

martinpopel:
without this option `t2t-decoder` crashes, see #397


lukaszkaiser:
Thanks! I'll try to remove this flag soon, but let's have it for now.

PR #412, opened 2017-11-11
Add option to profile during training

epurdyf:


rsepassi:
Have you tried running with the `--dbgprofile` flag? It should output a trace that can be viewed in Chrome.

epurdyf:
I hadn't noticed that. The ProfileContext profiler has slightly more useful output though, I think.

PR #419, opened 2017-11-14
Added a librispeech data generator.

wingsbr:
Added a generator for the librispeech datasets and included it in the supported generators in t2t-datagen.

Like the audio / TIMIT generator, this is dependent upon sox for WAV file generation.

### DIFF
--- a/tensor2tensor/data_generators/librispeech.py
+++ b/tensor2tensor/data_generators/librispeech.py
@@ -82,7 +82,7 @@ def _get_audio_data(filepath):
 class LibrispeechTextEncoder(text_encoder.TextEncoder):
 
   def encode(self, s):
-    return [ord[c] for c in s]
+    return [self._num_reserved_ids + ord(c) for c in s]
 
   def decode(self, ids):
     """Transform a sequence of int ids into a human-readable string.
@@ -97,7 +97,7 @@ class LibrispeechTextEncoder(text_encoder.TextEncoder):
       if 0 <= id_ < self._num_reserved_ids:
         decoded_ids.append(RESERVED_TOKENS[int(id_)])
       else:
-        decoded_ids.append(id_)
+        decoded_ids.append(id_ - self._num_reserved_ids)
     return "".join([chr(d) for d in decoded_ids])
  
 
@@ -199,7 +199,7 @@ class Librispeech(problem.Problem):
 
   @property
   def num_shards(self):
-    return 10
+    return 100
 
   @property
   def use_subword_tokenizer(self):
@@ -214,9 +214,9 @@ class Librispeech(problem.Problem):
     """If true, we only generate training data and hold out shards for dev."""
     return False
           
-  def feature_encoders(self, data_dir):
+  def feature_encoders(self, _):
     return {
-        "inputs": text_encoder.TextEncoder(), #None, #DoNothingEncoder(),
+        "inputs": text_encoder.TextEncoder(),
         "targets": LibrispeechTextEncoder(),
     }
     
@@ -233,8 +233,9 @@ class Librispeech(problem.Problem):
 
     
   def generator(self, data_dir, tmp_dir, training, eos_list=None, start_from=0, how_many=0):
-    eos_list = [1]
+    eos_list = [1] if eos_list is None else eos_list
     datasets = (_LIBRISPEECH_TRAIN_DATASETS if training else _LIBRISPEECH_TEST_DATASETS)
+    num_reserved_ids = self.feature_encoders(None)["targets"].num_reserved_ids
     i = 0
     for url, subdir in datasets:
       filename = os.path.basename(url)
@@ -260,7 +261,7 @@ class Librispeech(problem.Problem):
         i += 1
         audio_data, sample_count, sample_width, num_channels = _get_audio_data(
             media_file)
-        label = [ord(c) for c in text_data] + eos_list
+        label = [num_reserved_ids + ord(c) for c in text_data] + eos_list
         yield {
             "inputs": audio_data,
             "audio/channel_count": [num_channels],
--- a/tensor2tensor/data_generators/librispeech.py
+++ b/tensor2tensor/data_generators/librispeech.py
@@ -1,12 +1,20 @@
+from tensor2tensor.data_generators import problem
+from tensor2tensor.utils import registry
+from tensor2tensor.models import transformer
+from tensor2tensor.utils import modality
+from tensor2tensor.layers import common_layers
+from tensor2tensor.data_generators import text_encoder
+import random
+import tensorflow as tf
+import numpy as np
+from tensor2tensor.data_generators import generator_utils
 import os
 from subprocess import call
 import tarfile
 import wave
-import numpy as np
-import six
-from tensor2tensor.data_generators import generator_utils
+      
 
-_LIBRISPEECH_TRAIN_DATASETS = [
+'''_LIBRISPEECH_TRAIN_DATASETS = [
     [
         "http://www.openslr.org/resources/12/train-clean-100.tar.gz",  # pylint: disable=line-too-long
         "train-clean-100"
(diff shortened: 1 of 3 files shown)


mschonwe:
@vince62s -- I work with @wingsbr.  I was collaborating with Archy de Berker via gitter.  Looks like a co-worker of his (Majid) of ended up posting another LibriSpeech data generator (which I haven't looked at yet).

To your question about MFCC, I thought for the 'problem' it would be best to leave pre-processing/spectral featurization/MFCC up to user, to allow for experimentation.  We'll make sure to link up with Majid to avoid duplication if/when we code up the modality bottom to handle these transformations.

vince62s:
Thanks for your feedback. As it is now (and it was similar for WSJ) I have the impression it is trying to take frames directly as inputs, which is really disturbing ...

mschonwe:
Agreed.  It is an important TODO.

I haven't seen good results published from working from raw waveform.  However, there are many papers with good results starting from spectral features.  But even here there is a lot of room for experimentation (e.g., mel-scale, #of bins, window-size, etc.).

I think going to MFCC is over-engineering the features, and it would be better to let the NN to derive its own features from the spectrum.  In any case, different researchers, and different domains may warrant different approaches.  Hence, I would argue against pushing the encoding into the TFRecords.

mjlaali:
@mschonwe Sorry, I did not see you pull request before sending mine. 

Regarding MFCC, I suggest that let put the pre-processing in problem.Problem class.

mschonwe:
@mjlaali No worries - let's chat on Gitter to coordinate incorporating the signal processing.
Thankfully the ops required are now part of TF1.4 (https://www.tensorflow.org/api_docs/python/tf/contrib/signal).

wingsbr:
@zh794390558 Good idea.  I expanded the librispeech generator to include a problem and modality and registered those rather than using _SUPPORTED_PROBLEM_GENERATORS.

tensor2tensor/data_generators/librispeech.py
```diff
@@ -29,6 +37,18 @@
         "http://www.openslr.org/resources/12/dev-other.tar.gz",
         "dev-other"
     ],
+]'''
```
zh794390558: Why comment above code ?

tensor2tensor/data_generators/librispeech.py
```diff
@@ -69,41 +89,233 @@ def _get_audio_data(filepath):
...
-      for f in corpus_tar:
-        if not os.path.isfile(os.path.join(tmp_dir, f.name)):
-          members.append(f)
-      corpus_tar.extractall(tmp_dir, members=members)
+      
+
+class LibrispeechTextEncoder(text_encoder.TextEncoder):
+
+  def encode(self, s):
+    return [ord[c] for c in s]
```
zh794390558: should inlcued `self._num_reserved_ids `

zh794390558: ord[c] is wrong.

zh794390558: why not encode like self._num_reserved_ids + i ?

tensor2tensor/data_generators/librispeech.py
```diff
@@ -69,41 +89,233 @@ def _get_audio_data(filepath):
...
+  def feature_encoders(self, data_dir):
+    return {
+        "inputs": text_encoder.TextEncoder(), #None, #DoNothingEncoder(),
+        "targets": LibrispeechTextEncoder(),
+    }
+    
+  def example_reading_spec(self):
+    data_fields = {
+        "inputs": tf.VarLenFeature(tf.int64),
+        #"audio/channel_count": tf.FixedLenFeature([], tf.int64),
```
zh794390558: thie can be reserved！

tensor2tensor/data_generators/librispeech.py
```diff
@@ -69,41 +89,233 @@ def _get_audio_data(filepath):
...
+        #"audio/sample_count": tf.FixedLenFeature([], tf.int64),
+        #"audio/sample_width": tf.FixedLenFeature([], tf.int64),
+        "targets": tf.VarLenFeature(tf.int64),
+    }
+    data_items_to_decoders = None
+    return (data_fields, data_items_to_decoders)
+
+    
+  def generator(self, data_dir, tmp_dir, training, eos_list=None, start_from=0, how_many=0):
+    eos_list = [1]
```
zh794390558: not good.

tensor2tensor/data_generators/librispeech.py
```diff
@@ -69,41 +89,233 @@ def _get_audio_data(filepath):
...
+    else:
+      generator_utils.generate_dataset_and_shuffle(
+          self.generator(data_dir, tmp_dir, True), train_paths,
+          self.generator(data_dir, tmp_dir, False), dev_paths)
+      
+    
+  def hparams(self, defaults, unused_model_hparams):
+    p = defaults    
+    p.stop_at_eos = int(False)
+    p.input_modality = { "inputs": ("audio:librispeech_modality", None) }
```
zh794390558: registry.Modalities.AUDIO

tensor2tensor/data_generators/librispeech.py
```diff
@@ -69,41 +89,233 @@ def _get_audio_data(filepath):
...
+  @property
+  def input_space_id(self):
+    return problem.SpaceID.AUDIO_SPECTRAL
+
+  @property
+  def target_space_id(self):
+    return problem.SpaceID.EN_CHR
+
+  @property
+  def num_shards(self):
```
zh794390558: transformer_base_single_gpu, does this used?

wingsbr: I'm not sure I understand the question, but I used transformer_base_single_gpu as the basis for the hparams because that was what was referenced in all of the examples:

https://github.com/tensorflow/tensor2tensor/blob/master/docs/new_problem.md
https://github.com/tensorflow/tensor2tensor/blob/master/docs/walkthrough.md
https://github.com/tensorflow/tensor2tensor/blob/master/README.md

tensor2tensor/data_generators/librispeech.py
wingsbr: Ah, good catch, I'll fix that syntax.  Regarding, num_reserved_ids, I based it on the timit generator:

https://github.com/tensorflow/tensor2tensor/blob/master/tensor2tensor/data_generators/audio.py#L150

which doesn't offset for num_reserved_ids, but you're right that it makes sense, so I will do so here.

tensor2tensor/data_generators/librispeech.py
wingsbr: Good catch, I had meant to fix that but forgot.  Doing so now.

tensor2tensor/data_generators/librispeech.py
wingsbr: Wouldn't that result in the base Audio modality being used, and bypass the custom signal processing added to `bottom()`?

tensor2tensor/data_generators/librispeech.py
wingsbr: I'm sorry, I don't understand.  What are you suggesting?

mschonwe:
@zh794390558 most of our work has been on Listen, Attend and Spell with various enhancements.  We were about to start working on implementing arxiv.org/pdf/1610.03022v1.pdf which, in part, uses convolutions rather than the pblstm of LAS.

We wanted to give the 'all you need is attention' transformer model a try, and utilize the framework extensions that t2t offers.  So far we haven't gotten the transformer model to do much more than learn the LM from the labels.

If the transformer model isn't viable for this task, perhaps we can collaborate implementing a t2t Problem for convolutional+rnn model (like 1610.03022v1).

zh794390558:
@mschonwe Sorry for later response. I work on Mandarin, also use the Listen, Attend and Spell model, now 
 want to try some tricks in T2T which successfully used in NMT, including 'All you need is attention'.  I think 'convolutional+rnn model' maybe a good attempt.

PR #422, opened 2017-11-14
Completing the CNN/Dailymail summarization pipeline

urvashik:
This commit adds the following features:
- Raw test data generation for the train, dev and test splits for the cnn/dailymail data. This allows for evaluating the performance of the model in a way that is comparable to other models

- A script to compute rouge scores using the official pyrouge distribution.
   - Ensure that you have pyrouge setup on your machine correctly (Will add a pyrouge test script in the future)
   - The script outputs warnings for empty model predictions. This is because the model did not generate a summary for that example. These were not suppressed since its a useful data point to have.

- A bash script that first runs the moses tokenizer on predictions and targets and then computes the rouge score using the aforementioned pyrouge based script.

- Just want to explicitly point out the changes made to `utils/decoding.py` in removing the extra space while generating decodes/targets files, since it was never merged before. Not sure if that might break other things.

### DIFF
--- a/tensor2tensor/data_generators/cnn_dailymail.py
+++ b/tensor2tensor/data_generators/cnn_dailymail.py
@@ -19,9 +19,10 @@ from __future__ import absolute_import
 from __future__ import division
 from __future__ import print_function
 
-import hashlib
+import io
 import os
 import tarfile
+import hashlib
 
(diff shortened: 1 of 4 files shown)


urvashik:
I signed it

PR #428, opened 2017-11-17
Cycle_gan Updated

nimaous:


nimaous:
I signed it

nimaous:
I removed .DS_Store binary files and I think it should be ok now. please let me know if there is any problem

nimaous:
I remove the binary files. I think now it is ok. please let me know if there is still any problem.

nimaous:
is it ok now?

PR #437, opened 2017-11-22
RNN baselines (requires TF 1.4)

fstahlberg:
This PR adds the [TF NMT tutorial](https://github.com/tensorflow/nmt) to the subdirectory tensor2tensor/models/tfnmt and integrates it into the T2T infrastructure.

Note that some of the models (for example the GNMT encoder) require TensorFlow 1.4.


fstahlberg:
This PR contains the entire code base of the tutorial - We don't need the entire code base, but I tried to keep the modifications to it at a minimum to make sure that we can still pull upstream commits without causing too many conflicts.

lukaszkaiser:
Sorry, but I think that importing the whole repo (with docs and LICENCE and everything) is too much! Please, could you just import the models? I think that's all we need!

fstahlberg:
Okay, I've removed all unnecessary files. I could shrink it even more by deleting the unused bits in the existing files, but this would ultimately reduce the chance to sync with the upstream repository. If having a slim code base is more important I'm happy to do that.

zh794390558:
@fstahlberg This PR does not using T2T frame.

fstahlberg:
Not sure if I understand. It adds a stripped down version of the TF-NMT tutorial to  tensor2tensor/models/tfnmt/. In there I tried to be as close to the original tutorial as possible to make it possible to update it once the tutorial is updated. The adaptor code to T2T is in tensor2tensor/models/tfnmt_model.py

zh794390558:
@fstahlberg I mean these model does not include a Problem, Modality.  The t2t input pipeline does not have input length info, which maybe not able to used in BiLSTM model.

fstahlberg:
It works eg. with the translate_ende_wmt32k problem. You are right that t2t does not expose the sequence lengths, but this may be better addressed in a separate PR (it applies to the current lstm model too). For now, similarly to the current lstm model, the sequence lengths are derived from the position of PAD symbols in the raw features.

zh794390558:
@fstahlberg https://github.com/tensorflow/tensorflow/issues/5588  input length be optional now.

rsepassi:
Could you add a test to verify that this all works? Specifically one that exercises `TFNmt`, the subclass of `T2TModel`.

fstahlberg:
This PR needs to be updated for the recent changes in the T2T architecture. I'll close it for now, and reopen it (with tests) once I have time again to work on it.

Issue #432, opened 2017-11-21
Logging with time


Gldkslfmsd:
Hello,
I miss time in logs. Only few log messages have current time inside, but I would appreciate if all of them had it. I could use it for estimating e.g. training time etc.

Would it be possible if instead of 
```
INFO:tensorflow:Importing user module t2t_usr from path
```
there would be

```
[10/20/17 18:48:01 INFO]:tensorflow:Importing user module t2t_usr from path
```
? 

Thank you!

martinpopel:
I see, you have figured out a way: https://github.com/Gldkslfmsd/tensor2tensor/commit/eae48b7a92591a8dde5fdb084f0c8ca6b7dc3df9
Couldn't the code be moved to t2t_usr_dir? Or maybe make a hook, which can be re-defined in t2t_usr_dir.
This way users could use any Python code to setup logging without modifying the t2t2 codebase and it would be more versatile than hypothetical options `--log_format` etc.
It would also solve the problem of @colmantse (mentioned on Gitter) who wants to redirect the logs to a file (but for some reason does not want to use standard redirect `t2t-trainer ... 2> my.log`).

colmantse:
thank you martinpopel for the pointer here.

colmantse:
I am trying to log info from log_device_placement , however, I cannot really tell what levels they are at, logging.info doesn't work. but I don't see they are error, warning, fatal.. etc... any clue?

Gldkslfmsd:
@martinpopel I think this time logging feature should be default, and not "add this line to your t2t_usr if you want it", that would be inconvenient. 
Of course it's possible to setup default logging format and make it configurable in t2t_usr, but that would work only for the logs after t2t_usr loading. (That's everything but the first one.) It's also possible to add second log stream handler, so it can log to a file and stderr at once.

I'm not much familiar with the core architecture now, maybe there's a better place for my _set_time_logging function...

rsepassi:
We'd welcome a PR that did something like look for a method defined in the t2t_usr_dir module, e.g. `initialize()` that could do things like modify logging. Don't want to make this particular one the default because we have our own logging schemes internally and I imagine others do as well.

rsepassi:
Copied from a PR comment:

[...]

PR #440, opened 2017-11-24
Logging with time #432

Gldkslfmsd:
Now it includes a timestamp to every log message.
The travis icon change should be reverted.


Gldkslfmsd:
I signed it!

2017-11-24 12:39 GMT+01:00 googlebot <notifications@github.com>:

PR #442, opened 2017-11-26
logging with time #432

Gldkslfmsd:
#432


rsepassi:
As mentioned in #432, we don't want to make this the default. However, what I can do is make these binaries importable (instead of pure scripts as they are now) so that you can wrap them yourself (e.g. my_t2t_decoder.py where main does some setup/modifications and then calls `t2t_decoder.main(argv)`). I'll try to have that for the next release.

PR #478, opened 2017-12-19
SymbolModality bottom_simple improvements

v-shmyhlo:
I think this change better describes intention of the code. Also a bit of code reuse.


v-shmyhlo:
I signed it!

rsepassi:
Thanks for this refactor @v-shmyhlo. We actually have a significant refactor coming out in the next few days, so `model_builder` will actually be completely gone and the model_fn code will move into `T2TModel`. If after the refactor you think it could still use some cleanup, please feel free to submit a PR. Very sorry that not communicating that plan led to duplicated effort.

v-shmyhlo:
Thank you for this information @rsepassi, i am actually very glad that you guys are doing this refactoring, right now it is pretty hard for me to navigate through model_builder source code, looking forward to see your improvements.

PR #486, opened 2017-12-22
Executable bit

martinpopel:
Reintroducing the +x bit, which somehow got lost in some previous version.


rsepassi:
@martinpopel What does adding the executable bit buy you? It seems that the Travis tests are fine. Is it just cosmetic?

The reason I ask is because these files are auto-generated when we export out of our internal repo out to GitHub and it's likely that the bit will be lost again unless I manually add it back every time. Yes, it'd be nice to have it automated somehow but that would require some eng work that I'd rather not do if I don't have to.

martinpopel:
When testing a new version (e.g. an unmerged PR) from git, I am used just to add the repo to PYTHONPATH and the `bin/` to PATH, but for this I need the scripts to be executable.
I had this in previous versions of my checkout, but now I am not sure if it was executable also on github.

I know that when installed with pip, the executable bit is set by pip for the installed scripts, so this does not effect most users. Still I think it is a common practice to store the Python scripts executable in git.

BTW: I understand you want to allow importing also the t2t-* scripts as libraries and thus you duplicate the code to *.py, but this seems like non-standard solution. The common solution is that the script is just a thin wrapper and all the important code (which could be useful for importing) is stored elsewhere (in standard *.py modules).

rsepassi:
Ah, I see.

Until I can figure out a way of getting things to play nice during our export out to GitHub, let's leave off the x bit. You can `pip install .` (or `pip install -e .` if you want it to pick up edits) from within the repo if you're testing something from a PR (which is what I do locally as well).

Agreed that copying the whole file is not ideal, and I'll see if I can clean it up.

PR #489, opened 2017-12-23
data_generators.wmt => data_generators.translate

iislucas:
`data_generators.wmt` seems to have become `data_generators.translate`.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tensor2tensor/489)


rsepassi:
Thanks!

PR #490, opened 2017-12-23
Add helpful (I think) instructions & bugfix

iislucas:
* Added some instructions for using pip to create editable install so you can locally test. 
* Also added commands to run the local test.
* Fixed bug with accidental `self` argument in hparams definition.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tensor2tensor/490)

### DIFF
--- a/docs/new_problem.md
+++ b/docs/new_problem.md
@@ -240,16 +240,38 @@ All hyperparamters inherit from `_default_hparams()` in `problem.py.` If you wou
 from tensor2tensor.models import transformer
 
 @registry.register_hparams
-def word2def_hparams(self):
+def word2def_hparams():
     hparams = transformer.transformer_base_single_gpu()  # Or whatever you'd like to build off.
     hparams.batch_size = 1024
     return hparams
 ```
 
+# Test the data generation
+If you are using a local checkout of tensor2tensor, you can install the local version in developer mode with `pip installe -e .` from the tensor2tensor directory.
+
+You will also need to add your new file to the [`all_problems.py`](https://github.com/tensorflow/tensor2tensor/blob/master/tensor2tensor/data_generators/all_problems.py) file too.
+
+You can test data generation with: 
+
+```bash
+PROBLEM=word2def
+DATA_DIR=$HOME/t2t_data
+TMP_DIR=/tmp/t2t_datagen
+mkdir -p $DATA_DIR $TMP_DIR
+
+t2t-datagen \
+  --data_dir=$DATA_DIR \
+  --tmp_dir=$TMP_DIR \
+  --problem=$PROBLEM
+```
+
+The `PROBLEM` variable is a lower-case version of the class that was regered with `@registry.register_problem()`.
+
 # Run the problem
 Now that we've gotten our problem set up, let's train a model and generate definitions. 
 
-We specify our problem name, the model, and hparams.
+To train, specify the problem name, the model, and hparams:
+
 ```bash
 PROBLEM=word2def
 MODEL=transformer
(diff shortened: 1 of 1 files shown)


docs/new_problem.md
```diff
@@ -240,16 +240,38 @@ All hyperparamters inherit from `_default_hparams()` in `problem.py.` If you wou
...
 @registry.register_hparams
-def word2def_hparams(self):
+def word2def_hparams():
     hparams = transformer.transformer_base_single_gpu()  # Or whatever you'd like to build off.
     hparams.batch_size = 1024
     return hparams
 ```
 
+# Test the data generation
+If you are using a local checkout of tensor2tensor, you can install the local version in developer mode with `pip installe -e .` from the tensor2tensor directory.
```
martinpopel: `pip install -e .` (delete extra `e`)
Also this instruction should follow the `--t2t_usr_dir` explanation (or link to https://github.com/tensorflow/tensor2tensor#adding-your-own-components) because in most cases usr_dir is enough and there is no need for local checkout.

docs/new_problem.md
```diff
@@ -240,16 +240,38 @@ All hyperparamters inherit from `_default_hparams()` in `problem.py.` If you wou
...
+
+You will also need to add your new file to the [`all_problems.py`](https://github.com/tensorflow/tensor2tensor/blob/master/tensor2tensor/data_generators/all_problems.py) file too.
```
martinpopel: Again, it should be clear that this is needed only at the moment when I want to contribute my dataset/problem to t2t. If I want to keep it private (e.g. because the resources are not publicly&automatically downloadable), the usr_dir solution is preferred.

docs/new_problem.md
```diff
@@ -240,16 +240,38 @@ All hyperparamters inherit from `_default_hparams()` in `problem.py.` If you wou
...
+TMP_DIR=/tmp/t2t_datagen
+mkdir -p $DATA_DIR $TMP_DIR
+
+t2t-datagen \
+  --data_dir=$DATA_DIR \
+  --tmp_dir=$TMP_DIR \
+  --problem=$PROBLEM
+```
+
+The `PROBLEM` variable is a lower-case version of the class that was regered with `@registry.register_problem()`.
```
martinpopel: regered -> registered
It is not just lower-casing, but also de-camel-casing https://github.com/tensorflow/tensor2tensor/blob/master/tensor2tensor/utils/registry.py#L83

docs/new_problem.md
iislucas: Nice; I've clarified that `--t2t_usr_dir` works for testing on a separate project, and `pip install -e .` is only for contributing directly to the tensor2tensor project, along with adding it to all_problems.py.

docs/new_problem.md
iislucas: done

docs/new_problem.md
iislucas: thanks, clarified.

docs/new_problem.md
```diff
@@ -240,16 +240,40 @@ All hyperparamters inherit from `_default_hparams()` in `problem.py.` If you wou
...
+  --data_dir=$DATA_DIR \
+  --tmp_dir=$TMP_DIR \
+  --problem=$PROBLEM
+```
+
+Where:
+*   `PROBLEM` is the name of the class that was registered with `@registry.register_problem()`, but converted from `CamelCase` to `snake_case`.
+*   `PATH_TO_YOUR_PROBLEM_DIR` is a path to the directory of your python problem file.
+
+If you are adding to the tensor2tensor repository, you can install the local cloned version in developer mode with `pip installe -e .` from the tensor2tensor directory. You will also need to add your new problem file to [`all_problems.py`](https://github.com/tensorflow/tensor2tensor/blob/master/tensor2tensor/data_generators/all_problems.py).
```
martinpopel: Great, now just `installe` -> `install` and maybe also
are adding -> plan to contribute
will also need -> can also
(because not all contributions involve adding a new problem definition)

iislucas: cool, done.

rsepassi:
Thank you!

PR #520, opened 2018-01-15
Change gym import to a late import

epurdyf:
This makes it so that users do not have to have gym installed in order to use tensor2tensor, unless they use a specific problem that depends on gym.

### DIFF
--- a/tensor2tensor/data_generators/gym.py
+++ b/tensor2tensor/data_generators/gym.py
@@ -23,8 +23,6 @@ import os
 
 # Dependency imports
 
-import gym
-
 from tensor2tensor.data_generators import generator_utils
 from tensor2tensor.data_generators import problem
 from tensor2tensor.utils import registry
@@ -47,6 +45,7 @@ class GymDiscreteProblem(problem.Problem):
 
   @property
   def env(self):
+    import gym
     if self._env is None:
       self._env = gym.make(self.env_name)
     return self._env


rsepassi:
If you `pip install tensor2tensor` it installs gym: https://github.com/tensorflow/tensor2tensor/blob/master/setup.py

cbockman:
The hope here was to minimize external dependencies, since gym is basically unused right now, and, even if it were, it wouldn't be needed except to execute gym-specific problems.

That said, understand if this is perceived dependency complexity overhead you'd rather not add.  We can just strike gym from our local fork--but we thought we'd throw up a PR in case it could be mainlined, as we figured we weren't the only ones who might appreciate a lowered set of dependencies.

rsepassi:
Why gym specifically? The gym package is tiny so it doesn't take a long
time to download and at least on my machine import is also very fast.
What's the cost that's motivated you to remove it?

cbockman:
Mostly just general software engineering hygiene--we're always trying to limit our code dependencies (you never know when something is going to bite you...).  Most/all of the other dependencies that trigger when core t2t loads are used either through t2t writ large, or we otherwise use them fairly heavily and are less of a concern (numpy, scipy, etc.).  

"gym" as a dependency only exists for a very specific problem that we aren't using, most of your external users probably aren't using, and which doesn't support any core t2t functionality.  

That said, we realize that this is a product heavily driven by google internal usage and we feel lucky to just ride that wave, so feel free to close this PR if the delta disrupts the internal workflow you have set in place.  :)

rsepassi:
Fair enough. I'd like to go with a slightly different approach, so I'm going to close this PR, but I'll have the gym dep out of the way in the next release. Thank you.

cbockman:
great, thanks @rsepassi !

martinpopel:
I didn't care about the `gym` extra dependency so much, but incidentally the newest version of `gym` has a [bug](https://github.com/openai/gym/issues/838) which prevents [Google Colab](https://goo.gl/wkHexj) to work.
Hopefully, they will fix `gym` soon. Nevertheless, it is a nice illustration why it is a good idea to reduce unnecessary dependencies.

PR #522, opened 2018-01-16
Show t2t-bleu usage

martinpopel:
Many researchers are still reporting approx_bleu and comparing it with proper BLEU reported in other papers.
We should guide them to not do this and minimize the risk of false-news papers, such as https://arxiv.org/abs/1712.09662.

### DIFF
--- a/README.md
+++ b/README.md
@@ -144,9 +144,12 @@ t2t-decoder \
   --hparams_set=$HPARAMS \
   --output_dir=$TRAIN_DIR \
   --decode_hparams="beam_size=$BEAM_SIZE,alpha=$ALPHA" \
-  --decode_from_file=$DECODE_FILE
+  --decode_from_file=$DECODE_FILE \
+  --decode_to_file=translation.en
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -144,9 +144,12 @@ t2t-decoder \
...
   --output_dir=$TRAIN_DIR \
   --decode_hparams="beam_size=$BEAM_SIZE,alpha=$ALPHA" \
-  --decode_from_file=$DECODE_FILE
+  --decode_from_file=$DECODE_FILE \
+  --decode_to_file=translation.en
 
-cat $DECODE_FILE.$MODEL.$HPARAMS.beam$BEAM_SIZE.alpha$ALPHA.decodes
+# Eval BLEU
+# (Always report proper BLEU in papers, not the internal approx_bleu.)
+t2t-bleu --translation=translation.en --reference=ref-translation.de
```
rsepassi: so where is this ref-translation.de file? this command won't actually work for a user out of the box, right? or am i missing something?

martinpopel: OK. Added `echo -e 'Hallo Welt\nAuf Wiedersehen Welt' > ref-translation.de`

rsepassi: perfect, thank you

Issue #523, opened 2018-01-17
Travis log too long


martinpopel:
Travis CI has 4MB limit on the log file, but current T2T tests exceed this limit.
As a result I cannot see the real cause why a [given build](https://travis-ci.org/tensorflow/tensor2tensor/jobs/329572628) failed.
Most of the log (if downloaded as a raw file) is full of download progress bars, e.g.
```
  Downloading mpmath-1.0.0.tar.gz (511kB)
[?25l
[K    2% |▋                               | 10kB 53.6MB/s eta 0:00:01
[K    4% |█▎                              | 20kB 41.3MB/s eta 0:00:01
[K    6% |██                              | 30kB 42.5MB/s eta 0:00:01
[K    8% |██▋                             | 40kB 41.3MB/s eta 0:00:01
[K    10% |███▏                            | 51kB 42.6MB/s eta 0:00:01
```
This is the first time I see this problem, but it seems that until a new pip version with [--progress-bar off](https://github.com/pypa/pip/pull/4194) option is released, we should use [a workaround](https://github.com/pypa/pip/issues/2756#issuecomment-268773883) `pip install package | cat && exit ${PIPESTATUS[0]}` or `set -o pipefail; pip install package | cat`.

rsepassi:
how about `pip install -q` to have the install be quiet?

rsepassi:
And `git clone -q`

rsepassi:
And thank you for spotting this.

martinpopel:
According to https://github.com/pypa/pip/issues/2756 if we provide the `--quiet` option, pip does not display exceptions, which is a bit unfortunate for Travis. That said, it would be still better than the current state.
And yes, `git clone -q` is a good idea.

Unfortunately, it seems Travis is now completely broken. See e.g. https://travis-ci.org/tensorflow/tensor2tensor/builds/329575693 (note that a given commit just changes README.md).
In https://github.com/tensorflow/tensor2tensor/pull/524 I have tried to silence the pip, but Travis again failed for other reasons.

rsepassi:
I think it’s fine not to see the pip exceptions in Travis. The focus of the
tests is not to test pip install. If a pip command fails, we'll see it and
can retry locally to see the error.

martinpopel:
[...]

PR #488, opened 2017-12-23
scripts for proper BLEU evaluation, batch translation and averaging

martinpopel:
`t2t-bleu` computes the "real" BLEU
(giving the same result as [sacréBLEU](https://github.com/awslabs/sockeye/tree/master/contrib/sacrebleu) with `--tokenization intl`
and as [mteval-v14.pl](https://github.com/moses-smt/mosesdecoder/blob/master/scripts/generic/mteval-v14.pl) with `--international-tokenization`).
It can be used in two ways:
* To evaluate one translated file: `t2t-bleu --translation=my-wmt13.de --reference=wmt13_deen.de`
* To evaluate all translations in a given directory (created e.g. by `t2t-translate-all`).

`t2t-translate-all` translates all checkpoints in a given directory (created by `t2t-trainer` or `t2t-avg-all`).
A custom command (e.g. SGE cluster wrapper) can be used instead of `t2t-decoder` for the translation.

`t2t-avg-all` for each checkpoint in a given directory
it averages it with the N preceding ones.

All three scripts wait a given number of minutes for new checkpoints
(produced by t2t-decoder, which can be run concurrently with these scripts).

This PR is an improved version of #436 (which was not fully merged).

PR #524, opened 2018-01-17
t2t-decoder with --checkpoint_path

martinpopel:
* `--checkpoint_path` allows `t2t-decoder` to load any model, not only the latest one
* `--keep_timestamp` sets the translated file's mtime according to the checkpoint (it is by default on, but active only with `--checkpoint_path` and `--decode_to_file`)
* prevent `tf.gfile.Glob` crashes due to concurrent filesystem edits

All this is needed for `t2t-translate-all` + `t2t-bleu` to work as expected (I forgot to add this commit to #488).
In addition I am trying to solve #523.


tensor2tensor/bin/t2t_translate_all.py
```diff
@@ -67,6 +67,7 @@
 
 
 def main(_):
+  FLAGS = flags.FLAGS
```
rsepassi: FLAGS defined above. rm

tensor2tensor/utils/bleu_hook.py
```diff
@@ -158,6 +159,7 @@ def property_chars(self, prefix):
     return "".join(six.unichr(x) for x in range(sys.maxunicode)
                    if unicodedata.category(six.unichr(x)).startswith(prefix))
 
+uregex = UnicodeRegex()
```
rsepassi: why is this top level now? would prefer to leave it where it is

tensor2tensor/utils/bleu_hook.py
```diff
@@ -209,7 +210,11 @@ def bleu_wrapper(ref_filename, hyp_filename, case_sensitive=False):
 def _read_stepfiles_list(path_prefix, path_suffix=".index", min_steps=0):
   """Return list of StepFiles sorted by step from files at path_prefix."""
   stepfiles = []
-  for filename in tf.gfile.Glob(path_prefix + "*-[0-9]*" + path_suffix):
+  # tf.gfile.Glob may crash with
+  # tensorflow.python.framework.errors_impl.NotFoundError:
+  # xy/model.ckpt-1130761_temp_9cb4cb0b0f5f4382b5ea947aadfb7a40; No such file or directory
+  # Let's use standard glob.glob instead.
+  for filename in glob.glob(path_prefix + '*-[0-9]*' + path_suffix):
```
rsepassi: would prefer tf.gfile.Glob. why is it crashing?

tensor2tensor/bin/t2t_translate_all.py
martinpopel: I can delete the current line 41 (I hope), but not this one.
FLAGS need to be a local variable because they are used few lines below -- https://github.com/martinpopel/tensor2tensor/blob/61005b0b1e063e3e26cb403cf6e8cdea3509c3b5/tensor2tensor/bin/t2t_translate_all.py#L94
Thanks to `**locals()`, the users can provide e.g `--decoder_command='qsub my-decoder.sh {params} --my_param={FLAGS.my_param}'` and I think it is more user-friendly than forcing them to write `--my_param={flags.FLAGS.my_param}`.

BTW: Last time someone deleted this line during an internal-Google refactor, but didn't noticed me within PR review as you do now. It was one of the reasons why `t2t_translate_all.py` didn't work (another reason was I was lazy to write a test).

tensor2tensor/utils/bleu_hook.py
martinpopel: Creating the `UnicodeRegex` instance is quite slow because I need to iterate twice over all Unicode characters.
It is really a bad idea to create a new instance for each line of the test set to be tokenized.
It caused the whole `t2t-bleu.py` to be about 10000 times slower (depending on the test set size).
As I note in my [commit log](https://github.com/tensorflow/tensor2tensor/pull/524/commits/747d4294ebe6e65e9e93649126ec3716fe300f3a), I would slightly prefer my original implementation with a class with static methods only (the class was there rather as a namespace).

tensor2tensor/bin/t2t_translate_all.py
rsepassi: I see. How about this?:

```
locals_and_flags = dict()
local_and_flags.update(FLAGS.__dict__)
locals_and_flags.update(locals())
...
.format(**locals_and_flags)
```

tensor2tensor/utils/bleu_hook.py
rsepassi: ah ok. thanks. fine as is then.

tensor2tensor/utils/bleu_hook.py
martinpopel: When saving a checkpoint, TensorFlow always creates a temp file first (which unfortunately matches the given wildcard pattern, but this is not important here I think).
This temporary file is deleted soon, but if this happens when `tf.gfile.Glob` is being called, it crashes.
It crashes within the `tf.gfile.Glob` (i.e. not within the for-cycle in my code) with tensorflow.python.framework.errors_impl.NotFoundError, which is difficult to handle (I could catch the exception and try the same call again, but that's too hacky).
According to my extensive tests, `glob.glob` does not have this non-deterministic bug.

tensor2tensor/bin/t2t_translate_all.py
martinpopel: OK. If you feel it's more elegant, I will do it this way (and test).

tensor2tensor/utils/bleu_hook.py
rsepassi: the benefit of tf.gfile is that it will support many underlying filesystems, including Google cloud storage and internal filesystems. glob won't know what to do with a path that starts with gs://.

would updating the pattern to 'ckpt-[0-9]*.index' work?
i'm thinking maybe not because it's probably failing on a IsDirectory call.

i'm inclined to advocate for a retrying_glob function to maintain portability across filesystems though i agree it's hacky.

martinpopel: I know the benefit of tf.gfile and this was the reason why I tried it first.
I would prefer a different solution:
* import tf.gfile wrapper once which overrides glob.glob, and keep glob.glob everywhere in the code
* make the wrapper detect special systems (`gs://`), but otherwise fallback to the standard (and reliable) `glob.glob`

If this is not acceptable, I will catch the exception and call again.
Even if the bug is rare, it is annoying when running many week-long experiments and some of the evaluation jobs crash.

> would updating the pattern to 'ckpt-[0-9]*.index' work

No. I use `stepfiles_iterator` also in `t2t-bleu`, where the files being iterated are not checkpoints, but translations of the test set, which I decided to have no extension (but `path_suffix` can be e.g. `.txt` as well). For this reason I call it `stepfiles` and not `checkpoints`.

rsepassi: Let's have the retrying tf.gfile.Glob with the exception catching. Monkey-patching a lib would be really confusing for people reading the code and not understanding why something is happening.

lkluo:
I encountered t2t-translate-all error:

tensor2tensor/bin/t2t_translate_all.py", line 99, in main
    ).format(**locals())
KeyError: 'FLAGS'

when I copies several lines of code up to line 99 from t2t_translate_all.py, and print out "params", turning out no errors.

martinpopel:
@rsepassi: I hope all comments from the review are now fixed.
@lkluo: Yes, this is one of the errors fixed by this PR.

tensor2tensor/utils/bleu_hook.py
```diff
@@ -205,11 +206,24 @@ def bleu_wrapper(ref_filename, hyp_filename, case_sensitive=False):
...
+  tensorflow.python.framework.errors_impl.NotFoundError:
+  xy/model.ckpt-1130761_temp_9cb4cb0b0f5f4382b5ea947aadfb7a40;
+  No such file or directory
+
+  Standard glob.glob does not have this bug, but does not hangle gs://...
+  So let's use tf.gfile.Glob twice to handle most concurrency problems.
+  """
+  try:
+    return tf.gfile.Glob(pattern)
+  except tensorflow.python.framework.errors_impl.NotFoundError:
```
rsepassi: tf.errors.NotFoundError

PR #527, opened 2018-01-19
Prevent use of continuous_train_and_eval with the ValidationMonitor

epurdyf:
This is a frustrating thing to debug, and an explicit failure would be much better for users.


rsepassi:
Don't want to have flag access here. Have added a warning if using the wrong schedule but have local_eval_frequency set. Thank you.

PR #552, opened 2018-02-02
v1.4.3

rsepassi:
Not quite ready. Need to get the latest PR merge in as well as a version bump.


tensor2tensor/models/transformer.py
```diff
@@ -802,7 +803,7 @@ def transformer_base_v1():
   hparams.initializer = "uniform_unit_scaling"
   hparams.weight_decay = 0.0
   hparams.optimizer_adam_beta1 = 0.9
-  hparams.optimizer_adam_beta2 = 0.98
+  hparams.optimizer_adam_beta2 = 0.997
```
martinpopel: Why changing it here in `_v1`? I thought the whole point of having `_v1` and `_v2` is to keep replicability (on best-effort basis).

rsepassi: Yes, agreed. Reverting. Thanks for catching it. It must have slipped through in review.

rsepassi:
Ready to merge

tensor2tensor/models/transformer.py
```diff
@@ -1209,29 +1183,53 @@ def transformer_relative_big():
...
+  # Adaptive batch sizes and sequence lengths are not supported on TPU.
+  # Instead, every batch has the same sequence length and the same batch size.
+  # Longer sequences are dropped and shorter ones are padded.
+  #
+  # It is therefore suggested to use a problem where examples have been combined
+  # to a longer length, e.g. the "_packed" problems.
+  #
+  # For problems with variable sequence lengths, this parameter controls the
+  # maximum sequence length.  Shorter sequences are dropped and longer ones
+  # are padded.
```
martinpopel: The other way round:-)

tensor2tensor/utils/t2t_model.py
```diff
@@ -284,20 +293,26 @@ def loss(self, logits, features):
 
   def optimize(self, loss, num_async_replicas=1):
     """Return a training op minimizing loss."""
+    tf.logging.info("Base learning rate: %f", self.hparams.learning_rate)
     lr = self.hparams.learning_rate * optimize.learning_rate_decay(self.hparams)
+    if num_async_replicas > 1:
+      tf.logging.info("Dividing learning rate by num_async_replicas: %d",
```
martinpopel: by **sqrt**(num_async_replicas)

rsepassi:
Thanks for catching these. Want to make a PR?

martinpopel:
It's too minor for a PR. Please fix it in your internal repo, I guess it's easier for you.

BTW, regarding the merging process:
* The commit logs in GitHub seem to be truncated, e.g. fa28c982c or e077e19f87e ending in "in" and "the" respectively.
* The first commit of each release always conflates the edit described in the log and a revert of all edits merged on GitHub since the last release, see e.g. b2c04c03b5c. This way it is difficult to see what was the actual edit. Also it is difficult to use `blame` and `bisect` in the GitHub commit history. Ideally, I would prefer if you rebase your commits on top of master before merging, but I guess you cannot do this for some reasons. In that case, would it be possible at least to include a dummy commit after each release with log "temporarily reverting commits done on GitHub" or something like this?

PR #577, opened 2018-02-12
v1.5.1

rsepassi:


rsepassi:
I'm trying to see if I can simply add tests against 1.4....

rsepassi:
Good to go once Travis passes.

tensor2tensor/bin/t2t_trainer.py
```diff
@@ -274,7 +274,9 @@ def save_metadata(hparams):
   # Save hparams as hparams.json
   hparams_fname = os.path.join(output_dir, "hparams.json")
   with tf.gfile.Open(hparams_fname, "w") as f:
-    f.write(hparams.to_json(indent=0, sort_keys=True))
+    # TODO(lukaszkaiser): use the first line once we require TF 1.5+.
```
martinpopel: TF 1.5+ is already required in [setup.py](https://github.com/tensorflow/tensor2tensor/blob/64d1f9666e1bb5e7813013ccc48b85cab77735dd/setup.py#L48-L49)

rsepassi:
Yeah we’ll update that. Cloud ML Engine only supports 1.4 and we’d like to
support launching on it.

PR #595, opened 2018-02-16
Move save_metadata function down to after create exp_fn

mschonwe:
Allows `hparams.json` file include hparams added by the problems.

### DIFF
--- a/tensor2tensor/bin/t2t_trainer.py
+++ b/tensor2tensor/bin/t2t_trainer.py
@@ -328,12 +328,12 @@ def main(argv):
   if argv:
     set_hparams_from_args(argv[1:])
   hparams = create_hparams()
-  if is_chief():
-    save_metadata(hparams)
 
   with maybe_cloud_tpu():
     exp_fn = create_experiment_fn()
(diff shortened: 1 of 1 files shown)


rsepassi:
Because the problem hparams are a nested object, the json will include some python string representation of that object, right? (vs. actual str/int/float values).

What's the motivation?

Problem's hparams don't change so i'm not seeing the benefit of logging them out on each run.

mschonwe:
@rsepassi Thank you for considering this. I find having _all_ the hparams relevant to a given run collected in one place is useful for troubleshooting, documentation, and understanding the components of the problem.  

There are a number of audio_ parameters (e.g. audio_sample_rate) for which having hparams logged would be useful.  Especially when people start to use their own datasets, it will be useful to validate params are set properly for the dataset being used. This presumes that problem's hparams can and will change, albeit infrequently.

There were not any problematic python string representations in the json when I tested with the librispeech_clean problem.  As far as I can tell the problem hparams get added to the hparam object just like any other param.

lukaszkaiser:
Merged, but please let us know if that solves it!

PR #600, opened 2018-02-19
Makes approx_vocab_size a property as was intended

wes-turner:
approx_vocab_size is treated as a property later on in the file in, for example, vocab_filename.


wes-turner:
approx_vocab_size is treated as a property later on in the file in, for example, vocab_filename.

PR #603, opened 2018-02-20
Makes approx_vocab_size a property

wes-turner:
One more approx_vocab_size that should be decorated with @property.  All the rest of the
instances of approx_vocab_size have this decoration.


wes-turner:
Closing/reopening because git is hard.

PR #608, opened 2018-02-21
fix txt_line_iterator to not stop at empty line

bwilbertz:
In the text problem re-factor commit [`765d33`](https://github.com/tensorflow/tensor2tensor/commit/765d33bb6890cc7b8f194257d9c71226c26e0162) the behavior of the text iterator was changed in a way that it stops on the first empty line.

This breaks the ende translation, since the parallel corpus contains quite some empty lines and therefore was stopping after 235 lines of `news-commentary-v12.de-en.en`.

Here is a fix, which `strip()`'s only when yielding the result as it was in version `1.5.1`.


bwilbertz:
CLA now signed.

PR #617, opened 2018-02-26
Fixed use of name_scope by using named arguments whenever needed

### LINKED ISSUE #615: TF 1.6 will break some layers because name_scope's constructor args changed
I just tried to use tensor2tensor with TF 1.6, which causes an exception `TypeError: unhashable type 'list'`

This is caused in the usage of `tf.name_scope`, which has now wants arguments like `tf.name_scope(name, default_name=default_name, values=values)` (whereas before no `default_name` argument existed). T2T calls name scope as `tf.name_scope(name, [list of values])`. To fix, T2T should instead call it using named arguments i.e `tf.name_scope(name, values=[list of values])`

as a temp workaround, i've been patching `name_scope` (instead of patching each individual layer in T2T)

```python
import tensorflow as tf

# have to monkey patch bc tensor2tensor doesn't work with tf 1.6 :(
# because name_scope wants arguments as tf.name_scope(name, default_name, values)
# whereas before it was tf.name_scope(name, values)
class PatchedNameScope(tf.name_scope):
    
    def __init__(self, name, values=None, default_name=None):
        super().__init__(name, values=values, default_name=default_name)
        
    
tf.name_scope = PatchedNameScope
```

aman-tiwari:
Fixed the use of name_scope by passing the `values` argument as a named argument wherever it is used. This should fix #615.

### DIFF
--- a/tensor2tensor/data_generators/imagenet.py
+++ b/tensor2tensor/data_generators/imagenet.py
@@ -235,7 +235,7 @@ def distorted_bounding_box_crop(image,
   Returns:
     (cropped image `Tensor`, distorted bbox `Tensor`).
   """
-  with tf.name_scope(scope, "distorted_bounding_box_crop", [image, bbox]):
+  with tf.name_scope(scope, default_name="distorted_bounding_box_crop", values=[image, bbox]):
     # Each bounding box has shape [1, num_boxes, box coords] and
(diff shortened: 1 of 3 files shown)


aman-tiwari:
I signed it!

PR #620, opened 2018-02-26
v1.5.3

rsepassi:


rsepassi:
Good to go.

tensor2tensor/bin/t2t_trainer.py
```diff
@@ -65,7 +65,7 @@
   flags.DEFINE_string("output_dir", "", "Base output directory for run.")
   flags.DEFINE_string("schedule", "continuous_train_and_eval",
                       "Method of Experiment to run.")
-  flags.DEFINE_integer("eval_steps", 10000,
+  flags.DEFINE_integer("eval_steps", 100,
```
martinpopel: Why?

rsepassi: Try to reduce the number of log statements on error.

tensor2tensor/models/research/adafactor_experiments.py
```diff
@@ -0,0 +1,173 @@
...
+def afx_mimic_adam():
+  """Emulating Adam - should be very similar to afx_adam."""
+  hparams = afx_adam()
+  mimic_adam_with_adafactor(hparams)
+  return hparams
+
+
+@registry.register_hparams
+def afx_base():
+  """Baseline - no momentum, beta=0.999."""
```
martinpopel: beta2=0.999

tensor2tensor/bin/t2t_trainer.py
martinpopel: What about silencing the end-of-sequence exception? Or what about using 0="the whole dev set" as the default and measuring its size (number of batches) first?

rsepassi:
Have a way of measuring its size with all the filtering and bucketing and
everything that might be going on?

tensor2tensor/bin/t2t_trainer.py
martinpopel: > Have a way of measuring its size with all the filtering and bucketing and everything that might be going on?

Yes. Simply iterate through all the data before the first evaluation and remember how many steps were needed. I am no tf.Dataset expert, but it feels strange there is no way to iterate through all the date (one epoch) without this hack with an arbitrarily high eval_steps and a bunch of end-of-sequence exceptions.

PR #625, opened 2018-02-28
Fix PTB data generator.

johnglover:
Fixes PTB data generator, which seemed to get overlooked in a recent refactor.

* Adds `is_generate_per_split` property.
* Downloads corpus before returning generator in generate_samples. This is because `Text2TextProblem`'s `generate_encoded_samples` method expects a generator to be returned from `generate_samples`, but for `TOKEN` vocabularies the file must already exist before `get_or_create_vocab` is called (https://github.com/tensorflow/tensor2tensor/blob/master/tensor2tensor/data_generators/text_problems.py#L222).
* Moves code to download corpus out of generate_samples method (seemed a bit clearer).


rasmi:
@lukaszkaiser -- can you review/merge?

PR #670, opened 2018-03-27
Fixed small issues with ML Engine script.

rasmi:
This fixes the following issues:

- A typo in FLAGS which was looking for `cloud_ml_engine_master_type` rather than `cloud_mlengine_master_type`. 
- An error being raised based on API client dependency versions (see https://github.com/google/google-api-python-client/issues/299)
- Some clarification in the docs when using non-GPU workers.


rasmi:
@lukaszkaiser  -- please merge!

PR #668, opened 2018-03-26
v1.5.6

rsepassi:

PR #675, opened 2018-03-29
_beam_decode_slow should not be wrapped in variable_scope(name)

### LINKED ISSUE #674: transformer_relative decoding fails: "Key transformer/transformer/... not found"
I train `HPARAMS=transformer_relative_big` ([Self-Attention with Relative Position Representations](https://arxiv.org/abs/1803.02155)) and it goes OK,
but the decoding fails with
```
NotFoundError (see above for traceback): Key transformer/transformer/body/decoder/layer_0/encdec_attention/layer_prepostprocess/layer_norm/layer_norm_bias not found in checkpoint
         [[Node: save/RestoreV2_1 = RestoreV2[dtypes=[DT_FLOAT, DT_FLOAT, DT_FLOAT, DT_FLOAT, DT_FLOAT, ..., DT_FLOAT, DT_FLOAT, DT_FLOAT, DT_FLOAT, DT_FLOAT], _device="/job:localhost/replica:0/task:0/device:CPU:
0"](_arg_save/Const_0_0, save/RestoreV2_1/tensor_names, save/RestoreV2_1/shape_and_slices)]]
         [[Node: save/RestoreV2_1/_55 = _Recv[client_terminated=false, recv_device="/job:localhost/replica:0/task:0/device:GPU:0", send_device="/job:localhost/replica:0/task:0/device:CPU:0", send_device_incarnati
on=1, tensor_name="edge_62_save/RestoreV2_1", tensor_type=DT_FLOAT, _device="/job:localhost/replica:0/task:0/device:GPU:0"]()]]
```
The checkpoint contains a key `transformer/body/decoder/layer_0/encdec_attention/layer_prepostprocess/layer_norm/layer_norm_bias`
but the decoding is looking for a key starting with **transformer/transformer**, i.e. there is one extra "transformer" name space.
The key is the first of decoder keys, so I guess it is not important that it is layer_norm_bias.

Interestingly, when I decode with HPARAMS=transformer_big_single_gpu (instead of the correct HPARAMS=transformer_relative_big), the decoding does not fail, it just produces very low BLEU, as could be expected because the model was not trained to get input absolute-position embeddings and at the same time it is missing the relative-position embeddings.

I made sure I am using the same T2T version and hparams for training and decoding.
tensor2tensor==1.5.6 (yes, it is not released yet, I am using it from #668)
tensorflow==1.6.0

martinpopel:
unlike _fast_decode which must be wrapped so.
Fixes #674 and allows to decode with transformer_relative.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -58,12 +58,13 @@ script:
   - t2t-decoder --problems=$T2T_PROBLEM --data_dir=$T2T_DATA_DIR --model=transformer --hparams_set=transformer_tiny --output_dir=$T2T_TRAIN_DIR --decode_hparams='num_samples=10'
 
   # Export and query (on Python 2 only)
-  - if [[ "$TRAVIS_PYTHON_VERSION" == "2.7" ]] && [[ "$TF_VERSION" == "1.5.*"  ]]; then
-        t2t-exporter --problems=$T2T_PROBLEM --data_dir=$T2T_DATA_DIR --model=transformer --hparams_set=transformer_tiny --output_dir=$T2T_TRAIN_DIR;
-        pip install tensorflow-serving-api;
-        tensorflow_model_server --port=9000 --model_name=my_model --model_base_path=$T2T_TRAIN_DIR/export/Servo &
-        sleep 10;
-        t2t-query-server --problem=$T2T_PROBLEM --server=localhost:9000 --servable_name=my_model --data_dir=$T2T_DATA_DIR --inputs_once='1 0 1 0 1 0';
-    fi
+  # Bug: https://github.com/tensorflow/serving/issues/819
+  #- if [[ "$TRAVIS_PYTHON_VERSION" == "2.7" ]] && [[ "$TF_VERSION" == "1.6.*"  ]]; then
+  #      t2t-exporter --problems=$T2T_PROBLEM --data_dir=$T2T_DATA_DIR --model=transformer --hparams_set=transformer_tiny --output_dir=$T2T_TRAIN_DIR;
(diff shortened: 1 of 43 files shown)


rsepassi:
Really sorry to retract the PR for 1.5.6 @martinpopel. If you update this PR to just include the relevant commit we can get it in today or tomorrow.

rsepassi:
(Same for the other PR you sent - sorry again)

martinpopel:
This PR fixes 848836347424e8cbd11f96eeae1cda5f7508ec1b which was not in 1.5.5, so I cannot simply rebase it on 1.5.5.
I can do another PR which fixes the attention other than dot_product properly. Is this what you want?
(But usually if I edit a code which was meanwhile changed in the Google internal repo, my change is merged here on GitHub, but than got overwritten in the nearest release, so I wanted to prevent this.)

rsepassi:
Ah I see. Then it probably makes sense to wait until we release 1.5.6.

martinpopel:
Meanwhile, I did it in #678. Without this, the relative transformer decoding does not work, so it seems better to include it in 1.5.6, unless you plan to release 1.5.7 soon. (BTW: I wonder how you did the evaluation for the paper. Maybe it worked is some older version?)

pshaw:
Hi Martin, indeed evaluation for the paper was done with an older version of tensor2tensor prior to fast decoding becoming the default for the transformer, which seems to only support dot_product attention currently.

https://github.com/tensorflow/tensor2tensor/pull/668/commits/848836347424e8cbd11f96eeae1cda5f7508ec1b was intended to get this working again, but thanks for uncovering an issue with the fix.

PR #678, opened 2018-03-30
Fix transformer decoding when using attention other than dot_product.

### LINKED ISSUE #674: transformer_relative decoding fails: "Key transformer/transformer/... not found"

martinpopel:
Note that _beam_decode_slow should not be wrapped in variable_scope(name)
unlike _fast_decode which must be wrapped so.
Fixes #674 and allows to decode with transformer_relative.
This PR is an substitute for #675 (same, but rebased on v1.5.5).

### DIFF
--- a/tensor2tensor/models/transformer.py
+++ b/tensor2tensor/models/transformer.py
@@ -225,6 +225,13 @@ class Transformer(t2t_model.T2TModel):
               None if using greedy decoding (beam_size=1)
       }
     """
+    if self._hparams.self_attention_type != "dot_product":
+      # Caching is not guaranteed to work with attention types other than
+      # dot_product.
+      # TODO(petershaw): Support fast decoding when using relative
+      # position representations, i.e. "dot_product_relative" attention.
+      return self._beam_decode_slow(features, decode_length, beam_size,
+                                    top_beams, alpha)
     with tf.variable_scope(self.name):
       return self._fast_decode(
           features, decode_length, beam_size, top_beams, alpha)


rsepassi:
So why does this fix things? Where was the extra variable scope coming from? Seems odd.
I've also pinged Peter Shaw who authored this so hopefully he'll comment soon.

martinpopel:
@rsepassi: The `Transformer` class inherits from `t2t_model.T2TModel` including the method `_beam_decode_slow`.
When this method is called from `_beam_decode` in `T2TModel` no variable scope is used
https://github.com/tensorflow/tensor2tensor/blob/518eaba/tensor2tensor/utils/t2t_model.py#L527-L544
So I think the same approach should be used in the `Transformer` class.
Empirically, it helped me to get rid of the double "transformer/transformer" scope and thus fix the problem.

I don't know why the `_fast_decode` method needs the variable scope and `_beam_decode_slow` does not.
Well, `_fast_decode` uses `with tf.variable_scope("body")` so here it is obvious that should be inside variable scope `transformer`.
Maybe it would be more consistent if the `with tf.variable_scope(self.name)` was inside `_fast_decode`, but I don't fully understand the code, so I cannot comment on this.

pshaw:
Thanks Martin for uncovering this issue with the fix.
I'll leave it up to you both, but perhaps it would be better to apply on top of https://github.com/tensorflow/tensor2tensor/pull/668/commits/848836347424e8cbd11f96eeae1cda5f7508ec1b to get the corresponding unit test and comments, and make merging simpler?

martinpopel:
I did two PRs: this one on top of 1.5.5 and before that #675 on top of 1.5.6. Let @rsepassi decide which one to merge, I don't care. Feel free to ignore both and just fix the few lines yourselves.
Either way, the unit tests and comments should be kept.

rsepassi:
@pshaw, would you be able to verify that decoding isn't working with `transformer_relative` and make the necessary fixes?

PR #701, opened 2018-04-11
RL model-based experiment

blazej0:


tensor2tensor/layers/modalities.py
```diff
@@ -100,7 +100,6 @@ def bottom_simple(self, x, name, reuse):
         x = tf.squeeze(x, axis=3)
       while len(x.get_shape()) < 3:
         x = tf.expand_dims(x, axis=-1)
-
```
lukaszkaiser: This doesn't seem to be necessary, can you revert and not change this file at all (just to reduce the size of this large PR).

tensor2tensor/rl/envs/atari_wrappers.py
```diff
@@ -1,139 +0,0 @@
-# coding=utf-8
```
lukaszkaiser: Are we removing the whole file? If not, let's not remove the licence.

tensor2tensor/rl/envs/in_graph_batch_env.py
```diff
@@ -1,58 +1,7 @@
-# coding=utf-8
```
lukaszkaiser: Ditto.

tensor2tensor/rl/envs/py_func_batch_env.py
```diff
@@ -0,0 +1,169 @@
+# coding=utf-8
```
lukaszkaiser: And here.

tensor2tensor/utils/metrics.py
```diff
@@ -285,9 +285,9 @@ def create_evaluation_metrics(problems, model_hparams):
   def make_problem_specific_metric_fn(metric_fn, problem_idx, weights_fn):
     """Create a metric fn conditioned on problem_idx."""
 
-    def problem_metric_fn(predictions, features):
+    def problem_metric_fn(predictions, features, labels):
       """Metric fn."""
-      labels = features.get("targets", None)
+      #labels = features.get("targets", None)
```
lukaszkaiser: This will cause problems with current T2T models if I'm not mistaken, is it really needed?

tensor2tensor/utils/t2t_model.py
```diff
@@ -352,7 +352,7 @@ def top(self, body_output, features):
           "problem_hparams.target_modality is a dict.")
       return self._top_single(body_output, target_modality, features)
 
-  def _loss_single(self, logits, target_modality, features):
+  def _loss_single(self, logits, target_modality, feature):
```
lukaszkaiser: Can you explain this change? It looks a little strange, is it intended?

tensor2tensor/utils/t2t_model.py
```diff
@@ -387,7 +387,7 @@ def loss(self, logits, features):
       assert not isinstance(target_modality, dict), (
           "model_body must return a dictionary of logits when "
           "problem_hparams.target_modality is a dict.")
-      return self._loss_single(logits, target_modality, features)
+      return self._loss_single(logits, target_modality, features["targets"])
```
lukaszkaiser: And here?

tensor2tensor/layers/modalities.py
blazej0: Sure, sorry for this.

tensor2tensor/rl/envs/atari_wrappers.py
blazej0: This file will be deleted as a whole, but we are missing the licence in the other files.

tensor2tensor/utils/metrics.py
blazej0: This is similar to the change in `_loss_single` - it is not longer true, that labels are always associated with key `targets` - they can be associated with different key. For this reason, we let the caller choose which metric she is interested (this happens in line 1030 and 1032 of t2t_model.py).
Again, I might be missing some context of t2t, so please correct me, if I'm wrong.

PR #702, opened 2018-04-11
Multiple input features for transformer model

franckbrl:
Hello,

Here is a proposition to add a translate problem with multiple source features for the transformer model. This is similar to [input features for RNN-based models](http://www.aclweb.org/anthology/W16-2209).

Additionally to input words, the problem can take an arbitrary number of features. A ready-to-use example is provided for English-to-French translation, with English PoS-tags and lemmas, in `tensor2tensor/data_generators/translate_enfr_src_feat.py`. Quick experiments with nearly no best checkpoint search and no averaging were performed and gave a BLEU of 39.76. In comparison, a basic model with only words as inputs (and +4M parameters) gave 39.14, which seems rather consistent with the results in the paper cited above.

The problem, model and input modalities are all defined in `tensor2tensor/data_generators/translate_source_features.py`.

This pull request also contains a modification of `tensor2tensor/models/transformer.py`, adding a specific hidden size for the encoder (`enc_hidden_size`), which can now be different from the decoder. Since all input features of arbitrary sizes are concatenated, the resulting encoder size has to be larger than `hidden_size`.

Specifically for inference, `tensor2tensor/bin/t2t_decoder.py` was added a new flag for the input file containing source features. Finally, `tensor2tensor/utils/decoding.py` was modified to pre-process all input features.

There is one issue I could not solve. The decoding time is much slower with the source feature model: `1m19.265s`, against `0m49.117s` for basic model, on the provided test set (measured with unix time). I probably missed something in `tensor2tensor/utils/decoding.py` and I would be grateful if anyone could help with this.


franckbrl:
I signed it!

franckbrl:
I signed it!

PR #611, opened 2018-02-23
en-vi: add IWSLT'15 English-Vietnamese as new problem

stefan-it:
Hi,

with this PR a new problem is added: English-Vietnamese using the IWSLT'15 dataset from the [Stanford NLP group](https://nlp.stanford.edu/projects/nmt/).

I trained an English to Vietnamese model for 125k steps on a NVIDIA GTX 1060. Here are some nice comparisons of BLEU score on the `tst2013` test set. Other BLEU scores are taken from the [Towards Neural Phrase-based Machine Translation](https://arxiv.org/abs/1706.05565) paper.

| Model                                                                                                 | BLEU (Beam Search)
| ----------------------------------------------------------------------------------------------------- | ------------------
| [Luong & Manning (2015)](https://nlp.stanford.edu/pubs/luong-manning-iwslt15.pdf)                     | 23.30
| Sequence-to-sequence model with attention                                                             | 26.10
| Neural Phrase-based Machine Translation [Huang et. al. (2017)](https://arxiv.org/abs/1706.05565)      | 27.69
| Neural Phrase-based Machine Translation + LM [Huang et. al. (2017)](https://arxiv.org/abs/1706.05565) | 28.07
| Transformer (Base)                                                                                    | **28.12** (cased)
| Transformer (Base)                                                                                    | **28.97** (uncased)

PR #703, opened 2018-04-11
Multiple input features for transformer model

franckbrl:
Hello,

Here is a proposition to add a translate problem with multiple source features for the transformer model. This is similar to [input features for RNN-based models](http://www.aclweb.org/anthology/W16-2209).

Additionally to input words, the problem can take an arbitrary number of features. A ready-to-use example is provided for English-to-French translation, with English PoS-tags and lemmas, in `tensor2tensor/data_generators/translate_enfr_src_feat.py`. Quick experiments with nearly no best checkpoint search and no averaging were performed and gave a BLEU of 39.76. In comparison, a basic model with only words as inputs (and +4M parameters) gave 39.14, which seems rather consistent with the results in the paper cited above.

The problem, model and input modalities are all defined in `tensor2tensor/data_generators/translate_source_features.py`.

This pull request also contains a modification of `tensor2tensor/models/transformer.py`, adding a specific hidden size for the encoder (`enc_hidden_size`), which can now be different from the decoder. Since all input features of arbitrary sizes are concatenated, the resulting encoder size has to be larger than `hidden_size`.

Specifically for inference, `tensor2tensor/bin/t2t_decoder.py` was added a new flag for the input file containing source features. Finally, `tensor2tensor/utils/decoding.py` was modified to pre-process all input features.

There is one issue I could not solve. The decoding time is much slower with the source feature model: `1m19.265s`, against `0m49.117s` for basic model, on the provided test set (measured with unix time). I probably missed something in `tensor2tensor/utils/decoding.py` and I would be grateful if anyone could help with this.


lukaszkaiser:
Please check the tests, is it just a fluke or did sth break?

franckbrl:
There was a small compatibility error with TF version 1.4. This is now fixed.

stefan-it:
@franckbrl Could you please rebase it on the latest `master` branch? Thanks :)

stefan-it:
@franckbrl This PR looks really great! Another question: what pipeline (tokenization, pos-tagging) did you use for creating the `*.sft` files? I would really like to extend this to other languages :)

franckbrl:
Thank you @stefan-it !

I added comments in `tensor2tensor/data_generators/translate_enfr_src_feat.py` about the pre-processing. I used the SMALL data from translate_enfr problem. It was cleaned, normalized and tokenized using Moses scripts. Lemmas and PoS-tags were obtained with Tree-Tagger.

I just did the rebasing on the latest `master` branch.

tensor2tensor/data_generators/translate_enfr_src_feat.py
```diff
@@ -0,0 +1,178 @@
...
+def transformer_sfeats_hparams():
+  # define initial transformer hparams here
+  hp = transformer.transformer_base()
+  #hp = transformer.transformer_big()
+
+  # feature vector size setting
+  # the order of the features is the same
+  # as in the source feature file. All
+  # sizes are separated by ':'
+  hp.add_hparam("source_feature_sizes", "16:56:8")
```
stefan-it: @franckbrl I think an explanation of the feature sizes would be great here. I guess `56` is the size of available tags? I'm not sure, but when I use the following command:

```bash
cat baseline-1M_train.clean.en.sft | tr ' ' '\n' | cut -d "|" -f 1 | sort | uniq | wc -l
```

The reported size is `58` instead of `56`.

franckbrl: `source_feature_size` sets the embedding size (number of units) of the feature. The vocabulary size is not set in advance, it is computed on the full training data. So here, the first feature embedding has a dimension 16, and the feature has a vocabulary of 58 PoS-tags.

franckbrl:
I just fixed an important issue with word tokenization during vocabulary generation.

franckbrl:
About the decoding speed issue, I ran the basic translate_enfr problem and decoding takes `1m11.904s`. Earlier, I observed faster decoding speed, but it was with a different tokenization. The source feature problem runs decoding in `1m20.870s`, which seems fair.

stefan-it:
@franckbrl Yes, there could be a big difference in BLEU, e.g. see this recent discussion in #611 or on [Gitter](https://gitter.im/tensor2tensor/Lobby?at=5acfe16c7c3a01610dd81b46).

And here's a nice [paper](https://arxiv.org/abs/1804.08771) of some more problems with BLEU.

I'm not sure, but I guess you can use `baseline-1M_train.clean.en.tok` as (tokenized) baseline for the `translate_enfr` experiment? 

Maybe @martinpopel can help with this issue :)

franckbrl:
Thanks @stefan-it ! I'm currently training a `translate_enfr` using `baseline-1M_train.clean.en.tok`, but it is being tokenized again in `data_generators/tokenize.py`. Is there an easy way to use you own tokenization and have t2t not tokenize again over it? For the source features in this PR, I split sentences by space and each token was STE-encoded using `SubwordTextEncoder.encode_without_tokenizing()`. It seems complicated to do this in `translate_enfr` and we would need to override a few methods of the problem.

franckbrl:
Hi @stefan-it ,

I just got results for the two following systems trained on `baseline-1M_train.clean.*.tok` for 300k updates:

- `translate_enfr`: 39.36 BLEU, `1m16.841s` decoding time
- `translate_enfr_src_feat`: 39.60 BLEU, `1m20.892s` decoding time

The PR was also just rebased. Now it's your call.

tensor2tensor/data_generators/translate_enfr_src_feat.py
stefan-it: Ah, ok, I understand it now :) What do you think about renaming it to `source_feature_hidden_sizes`? Thanks for your patience!

franckbrl: These are not `hidden` layers, since the embeddings given as input. We do have a specific hidden layer for the encoder, and its size is `enc_hidden_size`.

stefan-it: Sorry, my mistake - I updated my previous comment. `source_feature_embedding_sizes` would be ok?

franckbrl: This sounds good, and it can't be understood as vocabulary size anymore. Let's do it?

stefan-it: Would be great :)

franckbrl: Done!

franckbrl:
I just solved conflicts with the master branch again. Are you still interested in this PR?

stefan-it:
Only @lukaszkaiser and @rsepassi can accept pull requests here. I'm very interested in adopting this implementation for other language pairs :)

lukaszkaiser:
@franckbrl : I was looking at this PR and it is unluckily too large to be merged as-is. Changes to Transformer, even as simple as adding an hparam, are now a little harder to do as they propagate to a large code-base where Transformer is used in many ways (I mean internal Google code). So this part would need to be reverted. Additionally, there are a few points that need to be addressed in terms of general coding style:
* sfeats / SrcFeat etc. --> features (the style guide strongly suggests long names)
* VocabType --> use text_encoder.VOCAB_TYPE
* get_or_create_vocab --> why not text_problems.get_or_create_vocab?

In general: should we put some small features in text_problems.py to make it more widely accessible and reduce code duplication? In the current form we cannot accept this PR, but if you're happy to do it, I'll be eager to help with splitting it up and merging in multiple parts.

I guess the first step would be to create a minimal working version (maybe just one vocab type?) and add just a data generator. Should we try that?

franckbrl:
Hi @lukaszkaiser ,

We could work on this minimal working version, but we will probably need more than just adding a specific data generator for source features. At each position in the source sequence, all source embeddings have to be concatenated, which breaks the logics of one single `hidden_size` dimension.

One thing we could do is setting the source feature embedding dimensions to `hidden_size` and sum all these embeddings, just like we do with the word and positional embeddings. But I have no idea whether this will work (what if you sum 10 embeddings together?). This would be an experiment, and not a straightforward implementation of the paper I cited above.

Note that in this case, we would still need to modify the transformer architecture for the sum, but we would stay at the embedding level.

lukaszkaiser:
What you say (summing) sounds more easily doable indeed. I'd be happy to try it, the PRs would be best step-wise: first one just with the data, then the model changes (hopefully small) so we can separate the problems. Thanks!

franckbrl:
Alright, let's try that. A few questions before I get into it:
- The first thing is just a data iterator that handles multiple inputs and does nothing with it, is that right? And you recommend to work within `text_problems.py` as much as possible ?
- The second input would be a minimal modification of the architecture (embedding sums), based on what the new data generator produces.
- Do I just modify this PR or shall I make a new one at each step?

f-lng:
Any updates on this PR?

lukaszkaiser:
Yes: I think we should work within `text_problems.py` and make it in a series of new PRs. Sorry for being so late! Let me close this one, I'm making a promise to myself to be better with the new ones!

PR #705, opened 2018-04-12
grow topk reshape issue

2877992943:
without that change below, for language model problem, when use batch predict, all samples in the batch will end up with the same result after get topk in method grow_topk()

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,7 @@
 # Compiled python modules.
 *.pyc
+*DS_Store
+
 
 # Byte-compiled
 _pycache__/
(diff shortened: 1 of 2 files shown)


2877992943:
to be concretely, for example, with regard to language model problem, 2 samples in 1 batch like these[[208,93,56,40],[237,56,23,5]] ,at the first step of loop, when grow_topk ,correct topk ids should be  like [208,237], but turn out to be [208,208],this pull request will correct this.

lukaszkaiser:
Sorry, this caused problems with our transformer tests so I'll need to revert it in the next release. Could you tell me why you thought this should be transposed? What is the command you're having a problem with?

PR #730, opened 2018-04-20
is_not_finish and decode_length

2877992943:
there are 2 changes by customized need below:
1)is_not_finish(),sometimes need to decode as long as decode_length no matter whether meet EOS ,because there is still useful information after EOS
2)decode_length need to listen to features['decode_length'] which come from feed  and is different for each single sample to be predicted.
hope to merge,thanks.


yynil:
hello, @lukaszkaiser please help to merge these two modifications.
We found sometimes transformer will still produce meaningful outputs even when EOS is outputted already, it's pretty useful when we do entity extraction using transformer.
Then for the decode_length, we found transformer was sensitive for the input length. So the decode_length should be able to be adjusted for different inputs. That ability helps to make the model easier to be useful for online prediction services since the user inputs can't be assumed as some certain length text. 
These two modifications will not affect the default configuration since we make the trigger in the input features. 

Thanks a lot again!

tensor2tensor/models/transformer.py
```diff
@@ -388,7 +394,7 @@ def forced_logits():
         ret = tf.cond(
             tf.less(i, partial_targets_length), forced_logits, lambda: ret)
       return ret, cache
-
+    decode_length_decide_end=features['decode_length_decide_end'] if 'decode_length_decide_end' in features else False
```
rsepassi: I don't see the value of having this in features. How about getting it from `decode_hparams` and piping it through?

tensor2tensor/models/transformer.py
```diff
@@ -399,7 +405,8 @@ def forced_logits():
         beam_size=beam_size,
         top_beams=top_beams,
         alpha=alpha,
-        batch_size=batch_size)
+        batch_size=batch_size,
+        decode_length_decide_end=decode_length_decide_end)
```
rsepassi: Let's find a better name. How about `force_decode_length` that defaults to `False`?

tensor2tensor/models/transformer.py
```diff
@@ -418,7 +425,8 @@ def fast_decode(encoder_output,
                 top_beams=1,
                 alpha=1.0,
                 eos_id=beam_search.EOS_ID,
-                batch_size=None):
+                batch_size=None,
+                decode_length_decide_end=False):
```
rsepassi: Please document this (with the updated name) in the docstring below.

tensor2tensor/models/transformer.py
```diff
@@ -510,7 +518,8 @@ def inner_loop(i, finished, next_id, decoded_ids, cache, log_prob):
       return i + 1, finished, next_id, decoded_ids, cache, log_prob
 
     def is_not_finished(i, finished, *_):
-      return (i < decode_length) & tf.logical_not(tf.reduce_all(finished))
+      if decode_length_decide_end==True:return i < decode_length
```
rsepassi:
```
if ...:
  ...
else:
  ...
```

tensor2tensor/models/transformer.py
yynil: Since decode_hparams is used to build the decoding static graph, the decoding process can't adapt the decode length for different inputs.
More specifically, we build a raw Tensorflow session to do the continuously decoding online, so we can't afford to build the graph again and again. 
Actually my plan is to refactor our raw Tensorflow session decoder and send another pull request later.
Thanks a lot!

tensor2tensor/models/transformer.py
yynil: hi, @rsepassi  please review the recent change if it's ok to make the merge happen. Thanks a lot!

2877992943:
@lukaszkaiser  thank you for your suggestion, changs finished, please help merge.

2877992943:
@rsepassi  thank you for your suggestion, changs finished, please help merge.

rsepassi:
Approved and fixed a merge conflict, but waiting on Travis tests to pass before merging.

rsepassi:
Merged! Thanks!

PR #739, opened 2018-04-21
Add English-Estonian (WMT2018) as new problem

stefan-it:
Hi,

this PR adds English-Estonian as new problem. English-Estonian is new for [WMT18](http://www.statmt.org/wmt18/translation-task.html), so all training data from WMT site is used.

The development data is also taken from WMT. Unfortunately, the original development data includes SGML tags, so I stripped them using [`input-from-sgm.perl`](https://github.com/moses-smt/mosesdecoder/blob/master/scripts/ems/support/input-from-sgm.perl) from Moses.


rsepassi:
Wonderful, thanks!

Issue #444, opened 2017-11-27
[Tuning] Results are GPU-number and batch-size dependent


vince62s:
@lukaszkaiser 
This is to illustrate what I have discussed on gitter.

Working with WMT EN-FR, I have observed the following.
You can replicate the paper results with "transformer -base" with 4 GPU.
The BLEUapprox looks like this: (batch-size 4096, warmup step 6000)
![image](https://user-images.githubusercontent.com/15141326/33256270-a3795912-d351-11e7-83e4-ea941ba95dd5.png)

If I do the same on 3 GPU (batch size 4096 warmup step 8000), taking into account that I need to compare step 120K of 4GPU run vs 160K of the 3GPU run, I get this with a clear offset of 1 BLEU point.
The gap is never closed if we wait.
![image](https://user-images.githubusercontent.com/15141326/33256370-1618ac16-d352-11e7-83c1-cfdcfa19a9ee.png)

If I do the same on 2GPU, it's even lower, 1GPU same.

Also, I observed that it is very dependent on the batch size.
For instance if you lower to 3072 you don't get the same as with 4096
With 2048 even lower.

This makes impossible to replicate the Transformer BIG results since you can only fit a batch size of 2048 even on a GTX1080ti.

Hope this helps for better tuning.

martinpopel:
My experience is exactly the same. Lower batch size (either nominally lower with the same number of GPUs or effectively lower because of smaller number of GPUs) results in worse results, even if I train long enough to compensate for the lower batch.
I remember people working with Nematus or OpenNMT were surprised by this behavior of Transformer/T2T because their experience was that lower batch size leads to better results in the end (but slower of course, thus they sometimes start training with big batch and then switch to lower batch for fine-tuning).

mehmedes:
@vince62s you mentioned that the gap is never closed even if you _wait_ to compensate the batch size difference. 

Have you been able to compensate by decreasing learning rate when decreasing batch size ? As suggested [here,](https://arxiv.org/pdf/1706.02677.pdf) maybe it would make sense to define the learning rate by multiplying it by ```x``` and ```y``` if batch size changes nominally (the same number of GPUs ) by ```x``` and/or effectively by ```y``` (different number of GPUs)?

vince62s:
not sure, because adam / noam is supposed to be adaptive. all the tests I did (changing the lr) were not better.
All what I know is that there are 2 places where the number of replicas has an impact:
[...]

PR #750, opened 2018-04-26
Large training batches on limited GPU hardware

fstahlberg:
This PR adds a LargebatchAdam optimizer, which accumulates gradients over n batches and applies the Adam learning rule every n batches on the accumulated gradients. This makes it possible to arbitrarily increase the effective batch size / number of GPUs at cost of more training iterations. This technique is useful if the number of physical GPUs is limited or the GPU memory does not allow to increase the batch size any further. Large batch / multi-GPU training is often important for Transformer training as reported [here](https://github.com/tensorflow/tensor2tensor/issues/444#issuecomment-384102731).


stefan-it:
@fstahlberg Looks really cool! Could you provide some results of experiments, e.g. translation problem with BLEU score and a comparison with one vs. multiple GPUs?

fstahlberg:
Closing this in favour of #754 as some commits were submitted with an invalid e-mail address which caused problems with the Google CLA, and my attempts to amend the e-mail address in previous commits were not successful.

PR #754, opened 2018-04-29
Large training batches on limited GPU hardware

fstahlberg:
This PR adds a LargebatchAdam optimizer, which accumulates gradients over n batches and applies the Adam learning rule every n batches on the accumulated gradients. This makes it possible to arbitrarily increase the effective batch size / number of GPUs at cost of more training iterations. This technique is useful if the number of physical GPUs is limited or the GPU memory does not allow to increase the batch size any further. Large batch / multi-GPU training is often important for Transformer training as reported in #444 . See [Saunders et al., 2018](https://arxiv.org/abs/1805.00456) for more details.

See transformer_base_fake_gpu8 hparams set as an example.

This is a new version of the PR #750  which fixes issues with the Google CLA.


fstahlberg:
@vince62s Yes, I tried it with "delay factor" 8 - there should be no difference regarding GPU memory between 4 and 8. Did you use the same code?

@rsepassi thanks for the review, I'll work on it in the next few days.

rsepassi:
One more thought. Maybe the hparam name should be something like
multistep_opt_accumulate_steps.

tensor2tensor/models/transformer.py
```diff
@@ -1081,6 +1081,17 @@ def transformer_base_single_gpu():
...
 
 
+@registry.register_hparams
+def transformer_base_fake_gpu8():
+  """HParams for simulating 8 GPU transformer base model training
+  on a single GPU.
+  """
+  hparams = transformer_base()
+  hparams.optimizer = "LargebatchAdam"
+  hparams.add_hparam("fake_gpu_multiplier", 8)
```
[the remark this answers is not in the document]
fstahlberg: Done. I renamed it to optimizer_multistep_accumulate_steps for consistency with the optimizer_ada[m|factor]_* options

tensor2tensor/utils/largebatch_optimizer.py
```diff
@@ -0,0 +1,163 @@
...
+
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+# Dependency imports
+
+import tensorflow as tf
+from tensorflow.python.eager import context
+from tensorflow.python.framework import ops
```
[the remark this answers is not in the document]
fstahlberg: That was a legacy from using the AdamOptimizer as blueprint. Fixed now.

tensor2tensor/utils/largebatch_optimizer.py
```diff
@@ -0,0 +1,163 @@
...
+  def _resource_apply_dense(self, grad, var):
+    return self._apply_cond(
+      super(LargebatchAdamOptimizer, self)._resource_apply_dense, grad, var)
+
+  def _apply_sparse_shared(self, grad, var, indices, scatter_add):
+    return self._apply_cond(
+      super(LargebatchAdamOptimizer, self)._apply_sparse_shared, grad, var,
+      indices, scatter_add)
+
+  def _apply_sparse(self, grad, var):
```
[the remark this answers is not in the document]
fstahlberg: Sure that I can do that? The optimizer works even for sparse tensors, just not as efficient as it could be as I'm simply converting to dense and use _apply_dense

fstahlberg:
@rsepassi I've added a unit test, but I had to put an awkward version check to pass all tests since it doesn't work with TF < 1.6.

tensor2tensor/utils/learning_rate.py
```diff
@@ -70,6 +70,17 @@ def legacy_learning_rate_schedule(hparams):
   return ret * optimizer_correction * hparams.learning_rate
 
 
+def _global_step(hparams):
+  """Adjust global step if a multi-step optimizer is used."""
+  step = tf.to_float(tf.train.get_or_create_global_step())
+  multiplier = hparams.optimizer_multistep_accumulate_steps
+  if multiplier is not None and multiplier > 1:
```
rsepassi: if multiplier:

tensor2tensor/utils/learning_rate.py
```diff
@@ -70,6 +70,17 @@ def legacy_learning_rate_schedule(hparams):
...
+    step = step / tf.constant(multiplier, dtype=tf.float32)
```
rsepassi: tf.to_float(step) / tf.to_float(multiplier)

tensor2tensor/utils/multistep_optimizer_test.py
```diff
@@ -0,0 +1,123 @@
+# coding=utf-8
```
rsepassi: Thank you for adding this test, but I think the test should be a bit different and hopefully simpler:

Compare 2 things:
1. AdamOptimizer with batch size 32 for 1 step
2. MultistepAdamOptimizer with batch size 8 for 4 steps with n=4

We should see that the updates are identical (i.e. the variables end up in the same place)

fstahlberg: Hm am I not doing something like this? Just that I don't compare batch sizes but number of updates. For example, I compare the variables after

1. Adam: t steps with averaged gradients over n steps
2. MultistepAdam: t*n steps

for n=1..4 and t=1..3. Adam is implemented in numpy to avoid introducing dependencies from this test class to other parts of the TF code (like in the original Adam test)

fstahlberg: Alright np I'll change it

fstahlberg:
Test updated

rsepassi:
Looks great! Thanks so much for this contribution @fstahlberg. Really good work.

tensor2tensor/layers/common_hparams.py
```diff
@@ -64,6 +64,8 @@ def basic_params1():
       optimizer_adafactor_memory_exponent=0.8,
       optimizer_adafactor_clipping_threshold=1.0,
       optimizer_adafactor_multiply_by_parameter_scale=True,
+      # Number of accumulating steps for multi step optimizers.
+      optimizer_multistep_accumulate_steps=None,
```
senarvi: I think the default value should be 1 instead of None. Otherwise "NoneType takes no arguments" error will occur when parsing the value from --hparams flag.

nxphi47:
Hello, with hparams, how many `--train_steps` do we set in the training script to replicate exactly 100000 steps with 8 real gpus in the transformer paper?

Is it still --train_steps=100000 or --train_steps=800000 ?

fstahlberg:
@nxphi47 It is  --train_steps=800000

zxqchat:
@fstahlberg Thank you so much and I want to know How about the performence please ?

fstahlberg:
@zxqchat For example, if you set optimizer_multistep_accumulate_steps to 8 and multiply train_steps with 8 you get the same performance as with 8 times more GPUs.

PR #780, opened 2018-05-09
changed beam decoder stopping condition

mirkobronzi:
stopping condition of the beam decoder now works as follows:

-- case return only the first beam (stop_early=True): it will stop when the first beam is 'unbeatable' by the other alive beams. (basically the previous behavior)

-- case return all beams (stop_early=False): it will stop when all the first N beams are 'unbeatable' by the other alive beams.
Note that checking that the last beam (in the top N) is unbeatable implies that all the top N beams are unbeatable as well.

Note w.r.t. the test changes: I duplicated testNotGreedyBeamTwo so that there is now a version with stop_early=True and one with stop_early=False. The first checks only the first beam, the latter will check all the beams.

Other two tests changed as well: they were checking the beam output length (checking it was equal to INPUT_LENGTH + decode_length) - disabled these checks given that this is not true anymore.


f-lng:
Any update on this PR?

mirkobronzi:
@f-lng  - It seems to fail the pylint check.
I am not exacly sure why (it doesn't say) - any help appreciated there.

I think the code is working though - all the tests are passing.
(and I am using it since long time now)

Not sure who should take a look in accepting/commenting the MR...
maybe @rsepassi ?
(I think I saw your name in the decoder code comments)

martinpopel:
Pylint test fails because of https://travis-ci.org/tensorflow/tensor2tensor/jobs/392427948#L1727
`463 beam_search._is_finished: Unused argument 'finished_in_finished'`
I haven't tried to understand the code, but if the parameter is not needed it should be deleted as `_is_finished` is an underscore method.

mirkobronzi:
thanks @martinpopel - there are  other parameters there that are not used in that method.
(and I think they cannot be removed because they are inside a tf while loop)
But I noticed their names start with `unused_` , so, I renamed the `finished_in_finished` into `unused_finished_in_finished`, and it seems the pylint check is now passing.

There is now a problem with an autoencoder unit test:
https://travis-ci.org/tensorflow/tensor2tensor/jobs/397916033#L1489

but it seems a problem unrelated to this PR - see last comment here:
https://github.com/tensorflow/tensor2tensor/pull/899
(failing on the same test)

mirkobronzi:
closing for CLA problems - re-opened here: https://github.com/tensorflow/tensor2tensor/pull/965

PR #789, opened 2018-05-10
Enabling eval on real environment again.

blazej0:


tensor2tensor/rl/rl_trainer_lib.py
```diff
@@ -71,22 +73,33 @@ def define_train(hparams, environment_spec, event_dir):
       # Some environments reset environments automatically, when reached done
       # state. For them we shall record only every second episode.
       d = 2 if env_lambda().metadata.get("semantics.autoreset") else 1
-      eval_env_lambda = lambda: gym.wrappers.Monitor(  # pylint: disable=g-long-lambda
+      monitor_env_lambda = lambda: gym.wrappers.Monitor(  # pylint: disable=g-long-lambda
           env_lambda(), event_dir, video_callable=lambda i: i % d == 0)
       eval_env_lambda = (
```
blazej0: It is important to have the right hand side lambda (`monitor_env_lambda`) to have a different name than the left hand side. Otherwise it becomes an infinite recursion, see:

```
x = lambda: 1
x = lambda: x()
x()
```

tensor2tensor/rl/rl_trainer_lib.py
```diff
@@ -71,22 +73,33 @@ def define_train(hparams, environment_spec, event_dir):
...
     # TODO(blazej0): correct to the version below.
-    corrected = False
+    corrected = True
     eval_summary = tf.no_op()
     if corrected:
       _, eval_summary = collect.define_collect(
           policy_factory, eval_batch_env, hparams, eval_phase=True)
   return summary, eval_summary
 
+def parse_from_summary(summary, name):
```
blazej0: I'm not very proud about this function, are there more canonical methods for extracting some values from summaries?

lukaszkaiser:
This looks complicated. I back-ported the small stuff, but I'm thinking about getting evals just by tracking rewards in the data generator when generating real data. Will close this for now, let's revisit tomorrow.

PR #797, opened 2018-05-11
 Env simulation should start from random points.

blazej0:


tensor2tensor/data_generators/gym_utils.py
```diff
@@ -28,18 +28,17 @@
 class WarmupWrapper(gym.Wrapper):
   """Warmup wrapper."""
 
-  def __init__(self, env, num_frames, warm_up_examples=0):
+  def __init__(self, env, warm_up_examples=0):
```
lukaszkaiser: Please sync to head (Piotr already did this, I think).

blazej0: Done.

PR #819, opened 2018-05-18
beam decode slow

2877992943:
some minor change in beam_decode_slow


2877992943:
please help merge, just some obvious change thanks.

lukaszkaiser:
Sorry it took so long, but could you explain this change a bit more? What is it actually changing, why is it important? Maybe add a comment in code to say why that's needed and where it failed? Otherwise someone will change it back soon...

Issue #809, opened 2018-05-16
*bug* checkpoints are incompatible wrt training/learning_rate


martinpopel:
When training with `learning_rate_schedule=constant`, the checkpoints contain a key called `training/learning_rate`.
When training with `learning_rate_schedule=rsqrt_decay`, the checkpoints do not contain the key.
Thus one cannot continue training with a constant LR with a model pre-trained with rsqrt_decay, we get the `Key training/learning_rate not found in checkpoint` error.

Should the key be there and why?

tensor2tensor==1.6.0
tensorflow==1.6.0

yanboliang:
@martinpopel I sent a [PR](#842) to fix this issue, wish it can help you. Thanks.

lukaszkaiser:
After merging #842 this should be ok, right? Closing, but feel free to reopen if needed of course!

PR #842, opened 2018-05-29
Wrap constant learning rate as Tensor

yanboliang:
When training with ```learning_rate_schedule=constant```, ```learning_rate``` is float type and would be wrapped as TF Variable by TensorFlow at [here](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/contrib/layers/python/layers/optimizers.py#L179), thus to be saved at checkpoint. However, when training with ```learning_rate_schedule= rsqrt_decay``` and other schedule mechanism(non-constant), ```learning_rate``` would be normal Tensor rather than Variable, and will not be saved at checkpoint, that leads error like #809 .
As it's not necessary to save ```learning_rate``` at checkpoint file, we can restore it from ```global_step``` and learning rate schedule which are set by hparams.
This PR wrap ```learning_rate``` as Tensor for both constant and non-constant values, then TF will not save it in checkpoint file. If we load a pre-trained model, it doesn't complains errors like  #809 . I have tested this fix offline and it works well.

### DIFF
--- a/tensor2tensor/utils/learning_rate.py
+++ b/tensor2tensor/utils/learning_rate.py
@@ -44,17 +44,10 @@ def learning_rate_schedule(hparams):
   schedule_string = hparams.learning_rate_schedule
   names = schedule_string.split("*")
   names = [name.strip() for name in names if name.strip()]
-  lr_value = 1.0
+  ret = tf.constant(1.0)
   for name in names:
-    lr_value *= learning_rate_factor(name, step_num, hparams)
-
-  with tf.variable_scope("training"):
-    lr = tf.get_variable("learning_rate", [], trainable=False)
-    update_ops = tf.get_collection_ref(tf.GraphKeys.UPDATE_OPS)
-    assign_lr = tf.assign(lr, lr_value)
-    update_ops.append(assign_lr)
-
-  return lr
+    ret *= learning_rate_factor(name, step_num, hparams)
+  return ret
 
 
 def legacy_learning_rate_schedule(hparams):
--- a/tensor2tensor/utils/learning_rate.py
+++ b/tensor2tensor/utils/learning_rate.py
@@ -44,10 +44,17 @@ def learning_rate_schedule(hparams):
   schedule_string = hparams.learning_rate_schedule
   names = schedule_string.split("*")
   names = [name.strip() for name in names if name.strip()]
-  ret = 1.0
+  lr_value = 1.0
   for name in names:
-    ret *= learning_rate_factor(name, step_num, hparams)
-  return ret
+    lr_value *= learning_rate_factor(name, step_num, hparams)
+
+  with tf.variable_scope("training"):
+    lr = tf.get_variable("learning_rate", [], trainable=False)
+    update_ops = tf.get_collection_ref(tf.GraphKeys.UPDATE_OPS)
+    assign_lr = tf.assign(lr, lr_value)
+    update_ops.append(assign_lr)
+
+  return lr
 
 
 def legacy_learning_rate_schedule(hparams):


martinpopel:
Great, thanks for finding the cause of the #809 bug.

Do I understand it correctly that merging this PR means introducing incompatibility in the saved checkpoints? Checkpoints created with older T2T versions (except those with constant lr, which is very rare) will fail to load. I am afraid breaking the backward compatibility of checkpoints causes more harm than good for most users.

Why do we need to store the learning rate in the checkpoint? If the checkpoint is used for inference only, no learning rate is needed. If the checkpoint is used for further training, we specify* the learning rate (and schedule) anyway, so I see no use case for reading the lr from a checkpoint.
*) I think all models use `layers/common_hparams.py` where the default values for both `learning_rate_constant` and `learning_rate` are specified. These defaults may be overridden elsewhere, but it cannot happen that the LR is unspecified.

When continuing training a model with constant LR, we may actually want to change that constant to a different value. If I understand [the code](https://github.com/tensorflow/tensorflow/blob/694892d4c0fba6bd4322e943f8b1483b36f1ae99/tensorflow/contrib/layers/python/layers/optimizers.py#L179) correctly, it does not allow such changes - once loading a model from a checkpoint where `training/learning_rate` is specified, this value is used and one cannot override it by hparams. This seems wrong to me.

So in conclusion, instead of this PR I would suggest to never save the learning rate to checkpoints. I think it is a better solution and moreover it does not break the checkpoint compatibility.

yanboliang:
@martinpopel Thanks for your reply, please see my inline comments bellow.

> Do I understand it correctly that merging this PR means introducing incompatibility in the saved checkpoints? Checkpoints created with older T2T versions (except those with constant lr, which is very rare) will fail to load. I am afraid breaking the backward compatibility of checkpoints causes more harm than good for most users.

I think we can't avoid breaking change, as older T2T versions handle constant LR and non-constant LR differently, we need to handle them in a unified way. 

> Why do we need to store the learning rate in the checkpoint? If the checkpoint is used for inference only, no learning rate is needed. If the checkpoint is used for further training, we specify* the learning rate (and schedule) anyway, so I see no use case for reading the lr from a checkpoint.
*) I think all models use layers/common_hparams.py where the default values for both learning_rate_constant and learning_rate are specified. These defaults may be overridden elsewhere, but it cannot happen that the LR is unspecified.

Fair enough, I agree it's not a must to save ```learning_rate``` as we can recover it by ```gloabl_step``` and learning schedule. But we should handle constant LR and non-constant LR in a unified way: save or don't save in a consistent way. I updated my PR to not save ```learning_rate``` for all cases, which I think is better solution. WDYT?

> When continuing training a model with constant LR, we may actually want to change that constant to a different value. If I understand the code correctly, it does not allow such changes - once loading a model from a checkpoint where training/learning_rate is specified, this value is used and one cannot override it by hparams. This seems wrong to me.

In my current fix, if you load a checkpoint and set a new learning rate by hparams, it can be override. As there is no ```learning_rate``` in the checkpoint, the graph will be filled what you set by hparams.

martinpopel:
Yes, this fix is great.

If the T2T core developers agree, I would suggest to delete the option that learning rate can be a float: both in the [docstring](https://github.com/tensorflow/tensorflow/blob/694892d4c0fba6bd4322e943f8b1483b36f1ae99/tensorflow/contrib/layers/python/layers/optimizers.py#L99) and in the [code](https://github.com/tensorflow/tensorflow/blob/694892d4c0fba6bd4322e943f8b1483b36f1ae99/tensorflow/contrib/layers/python/layers/optimizers.py#L176-L184). This is to prevent the problems with storing learning rate in checkpoints in other parts of T2T code base or in future. Or at least substitute the `lr = vs.get_variable(...)` part with `lr = tf.constant(learning_rate). This can be done either in this PR or in a separate one (after this one is merged).

yanboliang:
@martinpopel I take your point. I'd prefer to do it in a separate PR, as the [code](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/contrib/layers/python/layers/optimizers.py#L179) is inside core TF (not T2T), which may be called by many other places. I will open a separate thread to discuss that change. Let's focus on the T2T part in this PR. Thanks.

martinpopel:
> as that part of code is inside core TF (not T2T)

Oh, I haven't noticed. So of course, it must be a separate PR (or rather first an issue to discuss it).
Strictly speaking forbidding the float as the `learning_rate` argument is a backward incompatible change because it was documented, so removing it may not be allowed before TF 2.0.0 according to semantic versioning (though I am not sure whether it applies to contrib as well). Using `lr = tf.constant(learning_rate)` may be allowed, but I am less sure if there is no use case for the original behavior.

Anyway, this PR is ready for merge, thanks again.

afrozenator:
Thanks for the PR @yanboliang -- I'll merge this in now! Thanks for contributing!

Issue #467, opened 2017-12-08
translate_enzh_wmt8k_rev didn't support


beichao1314:
Hi， everyone. Is there anyone who have train a Chinese to English transformer model? When running data generation, translate_enzh_wmt8k_rev didn't support.
```
beichao@ubuntu16:~/train/enzh_3k5_1119/t2t-zhen$ ~/anaconda2/bin/t2t-datagen --data_dir=./datagen/ --tmp_dir=./data --problem=translate_enzh_wmt8k_rev
Traceback (most recent call last):
  File "/home/beichao/anaconda2/bin/t2t-datagen", line 4, in <module>
    __import__('pkg_resources').run_script('tensor2tensor==1.3.2', 't2t-datagen')
  File "/home/beichao/anaconda2/lib/python2.7/site-packages/setuptools-27.2.0-py2.7.egg/pkg_resources/__init__.py", line 744, in run_script
  File "/home/beichao/anaconda2/lib/python2.7/site-packages/setuptools-27.2.0-py2.7.egg/pkg_resources/__init__.py", line 1499, in run_script
  File "/home/beichao/anaconda2/lib/python2.7/site-packages/tensor2tensor-1.3.2-py2.7.egg/EGG-INFO/scripts/t2t-datagen", line 212, in <module>
    tf.app.run()
  File "/home/beichao/anaconda2/lib/python2.7/site-packages/tensorflow/python/platform/app.py", line 48, in run
    _sys.exit(main(_sys.argv[:1] + flags_passthrough))
  File "/home/beichao/anaconda2/lib/python2.7/site-packages/tensor2tensor-1.3.2-py2.7.egg/EGG-INFO/scripts/t2t-datagen", line 159, in main
    raise ValueError(error_msg)
ValueError: You must specify one of the supported problems to generate data for:
  * algorithmic_addition_binary40
  * algorithmic_addition_decimal40
  * algorithmic_algebra_inverse
  * algorithmic_cipher_shift200
  * algorithmic_cipher_shift5
  * algorithmic_cipher_vigenere200
  * algorithmic_cipher_vigenere5
  * algorithmic_identity_binary40
  * algorithmic_identity_decimal40
  * algorithmic_multiplication_binary40
  * algorithmic_multiplication_decimal40
  * algorithmic_reverse_binary40
  * algorithmic_reverse_binary40_test
  * algorithmic_reverse_decimal40
  * algorithmic_reverse_nlplike32k
[...]

PR #863, opened 2018-06-10
Added note to download original problem data when using the  flag for…

ShahNewazKhan:
… translation data-sets
Added a note to perform translation problems with the  `_rev` flag as per the solution in https://github.com/tensorflow/tensor2tensor/issues/467


ShahNewazKhan:
I signed the CLA!

ShahNewazKhan:
Updated alternate email registered with the CLA as defined in the commit authors.

PR #877, opened 2018-06-19
Fix CMVN variance normalization

jarfo:
Variance normalization should be performed dividing by the square root of the variance, not the variance

### DIFF
--- a/tensor2tensor/data_generators/speech_recognition.py
+++ b/tensor2tensor/data_generators/speech_recognition.py
@@ -310,10 +310,10 @@ class SpeechRecognitionProblem(problem.Problem):
       assert fbank_size[0] == 1
 
       # This replaces CMVN estimation on data
-
+      var_epsilon = 1e-09
       mean = tf.reduce_mean(mel_fbanks, keepdims=True, axis=1)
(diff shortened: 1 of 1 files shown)


jarfo:
CLA Signed!

lukaszkaiser:
DId you train models with this change? Does it help?

jarfo:
Hi Lukasz
I have tried with the fisher_callhome_spanish task. The epsilon part was necessary to avoid NaNs. An the normalization part also helped the convergence on TPUs. We are still having slow convergence problems on GPUs, even with many GPUs, and I ran out of TPU credit.
I would be pleased to evaluate the impact on librispeech both on TPUs and GPUs, and to provide new models. But I would need first some Google Cloud credit for the TPU part. I have received no answer yet from the TensorFlow Research Cloud program.

      De: Lukasz Kaiser <notifications@github.com>
 Para: tensorflow/tensor2tensor <tensor2tensor@noreply.github.com> 
CC: Jose Fonollosa <jarfo@yahoo.com>; Author <author@noreply.github.com>
 Enviado: Viernes 22 de junio de 2018 0:38
 Asunto: Re: [tensorflow/tensor2tensor] Fix CMVN variance normalization (#877)
   
DId you train models with this change? Does it help?—

PR #890, opened 2018-06-25
Eval reimplmentations

piotrmilos:


tensor2tensor/rl/envs/simulated_batch_env.py
```diff
@@ -112,20 +114,22 @@ def __init__(self, hparams, length,
     _, self.action_shape, self.action_dtype = get_action_space(environment_spec)
 
     # TODO(lukaszkaiser): do this in a more cleaner way
+    # remove other_hparams
+    hparams = copy.copy(other_hparams)
     hparams.video_num_input_frames, hparams.video_num_target_frames = (
-        hparams.model_hparams.video_num_input_frames,
-        hparams.model_hparams.video_num_target_frames)
+        hparams.environment_spec.video_num_input_frames,
```
lukaszkaiser: Why not take it from model hparams?

piotrmilos: The idea is that environment_spec should have all data needed to instantiate SimulateBatchEnv. Otherwise this are getting really confusing. It is possible that my choice is not the best, for example may be we should have environment_spec.model field?

piotrmilos: My intention is that environment_spec has all information needed to instantiate SimulatedBatchEnv. Otherwise, I found it hard to understand what the code really does.

lukaszkaiser:
Sorry, travis is broken by errors coming from this PR. Please correct.

tensor2tensor/rl/collect.py
```diff
@@ -112,15 +120,20 @@ def define_collect(hparams, scope, eval_phase,
...
 
+    eval_phase = tf.convert_to_tensor(eval_phase)
     should_reset_var = tf.Variable(True, trainable=False)
-
     zeros_tensor = tf.zeros(len(batch_env))
 
+  if "force_beginning_resets" in hparams:
+    force_beginning_resets = hparams.force_beginning_resets
+  else:
+    force_beginning_resets = False
```
mbz: default value for force_beginning_resets is not consistent across the board. Can we switch to all True or all False?

tensor2tensor/rl/envs/simulated_batch_env.py
lukaszkaiser: But it breaks the stochastic test. At this point this code is impossible to merge as it breaks tests and you're blocking Mohammad since the whole loop stopped working after the refactor. So we need to try to merge this ASAP and check if it works, or just revert the whole refactor to go back to a working state.

lukaszkaiser:
Guys, the tests are failing making it impossible to merge and I'll be on a plane in a few hours and unable to help for a while. Should we just revert the whole refactor to get back to a working state? Mohammad, are you blocked by this?

PR #936, opened 2018-07-13
minor docs fix to adam

cbockman:
Removed note about re-scaling the lr with Adam.

I believe this is a legacy of the old way LR+Adam was handled and that the docs weren't updated when the code was updated.

Pushing a PR because this is confusing to read now.  :)  (Unless I'm wrong!)


afrozenator:
Thanks for the PR @cbockman !

PR #951, opened 2018-07-22
fixed bug with data_gen in style transfer problem

tlatkowski:
PR changes:
1. fixed bug with generating data for 'original english to modern one' problem,
2. added char-based style transfer problems,
3. added unit tests.

### DIFF
--- a/tensor2tensor/data_generators/style_transfer.py
+++ b/tensor2tensor/data_generators/style_transfer.py
@@ -57,7 +57,7 @@ _DEV_SHARDS = 1
 _SUBWORD_VOCAB_SIZE = 8000
 
 
-class StyleTransferProblem(text_problems.Text2TextProblem):
+class StyleTransferProblemShakespeare(text_problems.Text2TextProblem):
   """Base class for transferring styles problems"""
 
   @property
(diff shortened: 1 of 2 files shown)


tensor2tensor/data_generators/style_transfer.py
```diff
@@ -69,7 +69,10 @@ def source(self):
     raise NotImplementedError()
 
   def dataset_url(self, dataset_split):
-    raise NotImplementedError()
+    train = dataset_split == problem.DatasetSplit.TRAIN
```
afrozenator: Thanks for the PR @tlatkowski !

I can see why you pulled out these 4 lines into the base class, but the base class, as it is written now is should be agnostic about the actual dataset -- no? i.e. I liked your earlier style better, can we revert to that? It only costs us 4 lines.

Another solution would be to create yet another class, maybe `StyleTransferProblemShakespeare` (which is a base class of `StyleTransferProblem`) and have the `StyleTransferShakespeareToModern` and `StyleTransferModernToShakespeare` inherit from it -- however this is unnecessary complexity I feel.

WDYT?

afrozenator: Another thing struck me, just renaming `StyleTransferProblem` to `StyleTransferProblemShakespeare` -- will be good too, since the `generate_samples` maybe very specific as well.

LMK. Thanks!

tlatkowski:
Hi @afrozenator , thanks for your remarks, I agree that after the changes in code, the base class is not longer generic one. So, I renamed the base class to _StyleTransferProblemShakespeare_. Thanks!

afrozenator:
Thanks @tlatkowski !

PR #965, opened 2018-08-01
changed beam decoder stopping condition

mirkobronzi:
stopping condition of the beam decoder now works as follows:

-- case return only the first beam (stop_early=True): it will stop when the first beam is 'unbeatable' by the other alive beams. (basically the previous behavior)

-- case return all beams (stop_early=False): it will stop when all the first N beams are 'unbeatable' by the other alive beams.
Note that checking that the last beam (in the top N) is unbeatable implies that all the top N beams are unbeatable as well.

Note w.r.t. the test changes: I duplicated testNotGreedyBeamTwo so that there is now a version with stop_early=True and one with stop_early=False. The first checks only the first beam, the latter will check all the beams.

Other two tests changed as well: they were checking the beam output length (checking it was equal to INPUT_LENGTH + decode_length) - disabled these checks given that this is not true anymore.


KasparPeterson:
Speeds things up considerably for our team. Great job!
Can this be merged soon so that we can continue using the master branch version?

mirkobronzi:
happy to help @KasparPeterson 
again, it is my first MR, not sure how to proceed to get it accepted.

@rsepassi , any procedure that should be followed?

afrozenator:
Taking a look, sorry didn't spot this sooner - @mirkobronzi @KasparPeterson

afrozenator:
Many thanks @mirkobronzi -- this is awesome!

mirkobronzi:
great! thanks @afrozenator :)

shakier:
Thanks for taking care of this issue! Is this currently working? I install from the codes and tried the decoding with returning multiple beams, but it still takes long time. Is there anything I missed? Do I need to set any parameters? Thanks!

mirkobronzi:
@shakier : no, it should now use the new code when you ask for multiple beams in the results.

PR #970, opened 2018-08-04
Handling the cases where the target_modality is a dict

MostafaDehghani:
Handling the cases where the target_modality is a dict (mostly in functions used at inference time) in t2t_model and transformer.


lukaszkaiser:
Thanks Mostafa. Can you make sure the CLA passes? We need it to merge, unclear why it's not passing. Maybe try again?

MostafaDehghani:
Not sure what's wrong with the CLA. I disabled two-factor authentication and this might be the problem. Going to enable it and send a new pull request.

PR #971, opened 2018-08-05
from six.moves import range for Python 3

cclauss:
__xrange()__ was removed in Python 3 in favor of __range()__ so convert all instances of __xrange()__ to __six.moves.range()__ for compatibility across Python 2 and Python 3.

Discovered via #942 at https://travis-ci.org/tensorflow/tensor2tensor/jobs/404819029#L1509


afrozenator:
Thanks @cclauss !

PR #972, opened 2018-08-05
remove duplicate lines

xu-song:


afrozenator:
Thanks again @xu-song !

PR #974, opened 2018-08-07
Undeprecate exporting the model from the trainer

gcampax:
But make use of tf.estimator.Exporter

Exporting the model during training is useful, and it allows to export the best model, rather than the last checkpoint, which for some datasets and hyperparameters can make a difference.


tensor2tensor/data_generators/problem.py
```diff
@@ -933,6 +933,15 @@ def prepare_for_output(example):
 
     return dataset
 
+  @property
+  def export_asserts(self):
```
lukaszkaiser: Is this function strictly needed, or are you using it for some private problems? It's fine with me, but I'm worried that without any tests it'll just get deleted in the next code cleanup...

Also: the name should be "assets" and not "asserts", right?

gcampax: Yeah we're using it in some yet unreleased code that uses t2t with our own problems.

And yes thanks for spotting the typo, I'll fix it.

tensor2tensor/data_generators/problem.py
```diff
@@ -933,6 +933,15 @@ def prepare_for_output(example):
...
+    """Assets to export with the model.
+    
+    This property contains a dictionary of assets, such as vocabulary files,
+    that should be exported toghether with the model, or None if no assets
```
cbockman: toghether => together

Very happy to see this PR!

gcampax:
Is anything else needed before this can be merged?

tensor2tensor/data_generators/problem.py
```diff
@@ -931,6 +931,15 @@ def prepare_for_output(example):
 
     return dataset
 
+  @property
+  def export_assets(self):
```
afrozenator: Hi @gcampax - I still see the old name here, have you done another commit after taking @lukaszkaiser 's review into account?

Thanks

afrozenator: My bad, I misread his comments, I'll start the process of merging this in shortly!

Issue #968, opened 2018-08-03
Minor Typo (tf.logging.info)


repoloper:
May be not worth a ticket, but just wanted to bring it to attention. I guess it must be tf.logging.info as opposed to "into". 

https://github.com/tensorflow/tensor2tensor/blob/6a7ef7f79f56fdcb1b16ae76d7e61cb09033dc4f/tensor2tensor/utils/get_rouge.py#L49

afrozenator:
Thanks @repoloper -- would you like to contribute by sending a pull request :) 👍

repoloper:
Sure. Will do!

afrozenator:
Thanks @repoloper -- looking forward to it!

afrozenator:
Thanks @repoloper ! Marking as closed now.

PR #978, opened 2018-08-08
typo fixed from tf.logging.into to tf.logging.info

repoloper:
Fixed a typo mentioned in https://github.com/tensorflow/tensor2tensor/issues/968


repoloper:
I signed it.

PR #983, opened 2018-08-09
Handling the cases where the target_modality is a dict

MostafaDehghani:
Handling the cases where the target_modality is a dict (mostly in functions used at inference time) in t2t_model and transformer.


afrozenator:
@MostafaDehghani - Same problem as earlier, can you see it the git commit emails are the same as the CLA email (or maybe sign the CLA again with the git commit log email)?

Here is what I see when I see the commits, note that the avatars are different, implying different email addresses?

image

Thanks

MostafaDehghani:
Thanks @afrozenator. This is so weird! I added another email address that I have to the CLA agreement (although I'm pretty sure I didn't use it when I sent the pull request). Hope this solves the problem. I think I also need to sync my PR...

afrozenator:
@googlebot rescan

afrozenator:
@MostafaDehghani -- let's see if that does the trick

c.f. https://opensource.google.com/docs/cla/

afrozenator:
@googlebot rescan

MostafaDehghani:
@afrozenator I think I got what is the problem :) I'm going to close this PR and submit a fresh one soon :)

PR #987, opened 2018-08-13
Update universal_transformer_util.py fix bug TypeError

zxqchat:
TypeError: Tensor objects are not iterable when eager execution is not enabled. To iterate over this tensor use tf.map_fn.


zxqchat:
thank you for your reply,this is my first contribution , I have already signed the CLA.

PR #999, opened 2018-08-16
transformer_fast_decode, beam search: take an optional cache and return it

gcampax:
Some models, eg. semantic parsing models with copying mechanisms,
want to use the output of Transformer for multiple predictions.
One way to do so is to modify the symbols_to_logits_fn to generate
the additional predictions and save them in the cache dictionary.

To do so, though, fast_decode() must allow an externally supplied cache,
and must return it to the caller after the loop.


gcampax:
Note: I have added tests but I cannot run them because pytest claims "no test dir found testing here".

lukaszkaiser:
I'm terribly sorry that this PR got forgotten for such a long time! It's a great help indeed and was perfectly fine to merge from the start. Thanks again for doing this!

Issue #1004, opened 2018-08-17
a question about the code of universal transformer


cyvius96:
https://github.com/tensorflow/tensor2tensor/blob/2f8423a7daf39c549fa4f87d369d3ff95e719e6c/tensor2tensor/models/research/universal_transformer_util.py#L1207

Is this supposed to be "(previous_state * (1 - update_weights))" instead of "(previous_state * 1 - update_weights)"?

Thanks.

akikaaa:
It makes sense to be  `new_state = transformed_state * update_weights +previous_state` given that update_weights is just p_t (in some sense) and previous_state is just the sum of previous p_t*s_t

PR #1011, opened 2018-08-22
Fix to universal transformer update weights

cbockman:
This was identified (not by me) in https://github.com/tensorflow/tensor2tensor/issues/1004.

Putting in a PR to ensure this is fixed.

If this is not a correct fix, then the underlying code needs to be rejiggered for readability...


MostafaDehghani:
It's a correct fix! Thanks @cbockman  for the PR.

afrozenator:
Thanks @cyvius96 @cbockman  and @MostafaDehghani  -- will merge in shortly.

PR #1014, opened 2018-08-23
Clipwrapper

piotrmilos:


piotrmilos:
A tf style wrapper doing clipping of rewards

PR #1019, opened 2018-08-26
Autoencoder RL experiment fixes

koz4k:


tensor2tensor/data_generators/gym_problems.py
```diff
@@ -62,8 +62,8 @@ def standard_atari_env_spec(env):
 
 def standard_atari_ae_env_spec(env):
   """Parameters of environment specification."""
-  standard_wrappers = [[tf_atari_wrappers.StackAndSkipWrapper, {"skip": 4}],
-                       [tf_atari_wrappers.AutoencoderWrapper, {}]]
+  standard_wrappers = [[tf_atari_wrappers.AutoencoderWrapper, {}],
+                       [tf_atari_wrappers.StackAndSkipWrapper, {"skip": 4}]]
```
lukaszkaiser: We try to use just StackWrapper everywhere, could you change to that?

tensor2tensor/data_generators/gym_problems.py
```diff
@@ -591,7 +608,10 @@ class GymSimulatedDiscreteProblemAutoencoded(GymSimulatedDiscreteProblem):
 
   def get_environment_spec(self):
     env_spec = standard_atari_env_spec(self.env_name)
-    env_spec.wrappers = [[tf_atari_wrappers.IntToBitWrapper, {}]]
+    env_spec.wrappers = [
+        [tf_atari_wrappers.IntToBitWrapper, {}],
+        [tf_atari_wrappers.StackAndSkipWrapper, {"skip": 4}]
```
lukaszkaiser: Here too. Just StackWrapper {"history": 4} should work fine.

tensor2tensor/data_generators/gym_utils.py
```diff
@@ -29,7 +29,7 @@ def __init__(self, env, warm_up_examples=0, warmup_action=0):
     self.warm_up_examples = warm_up_examples
     self.warm_up_action = warmup_action
     self.observation_space = gym.spaces.Box(
-        low=0, high=255, shape=(210, 160, 3), dtype=np.uint8)
+        low=0, high=255, shape=(210, 160, 3), dtype=np.int32)
```
lukaszkaiser: Do we need int32? If these are uint8, why not keep them like this? If they're not, please add a comment why.

tensor2tensor/data_generators/gym_utils.py
```diff
@@ -68,7 +68,7 @@ def __init__(self, env, warm_up_examples=0,
       self.action_space = gym.spaces.Discrete(2)
     self.warm_up_examples = warm_up_examples
     self.observation_space = gym.spaces.Box(
-        low=0, high=255, shape=(210, 160, 3), dtype=np.uint8)
+        low=0, high=255, shape=(210, 160, 3), dtype=np.int32)
```
lukaszkaiser: And here.

tensor2tensor/data_generators/gym_utils.py
```diff
@@ -147,7 +147,7 @@ def __init__(self, env, warm_up_examples=0,
     self.warm_up_examples = warm_up_examples
     self.observation_space = gym.spaces.Box(low=0, high=255,
                                             shape=(210, 160, 3),
-                                            dtype=np.uint8)
+                                            dtype=np.int32)
```
lukaszkaiser: And here.

tensor2tensor/rl/envs/batch_env.py
```diff
@@ -43,11 +43,12 @@ def __init__(self, envs, blocking):
     """
     self._envs = envs
     self._blocking = blocking
-    observ_space = self._envs[0].observation_space
```
lukaszkaiser: If we're using only the dtype externally to this code block, please make only that part of the object to limit leakage. I.e.,

observ_space = self._envs[0].observation_space
self.observ_dtype = observ_space.dtype

tensor2tensor/rl/envs/tf_atari_wrappers.py
```diff
@@ -25,6 +25,7 @@
 from tensor2tensor.layers import discretization
 from tensor2tensor.models.research import autoencoders
 from tensor2tensor.rl.envs.in_graph_batch_env import InGraphBatchEnv
+from tensor2tensor.utils import trainer_lib
```
lukaszkaiser: This is a huge import to otherwise mostly self-standing file. Let's do the other changes in one PR first and then consider this.

tensor2tensor/rl/envs/tf_atari_wrappers.py
```diff
@@ -200,19 +207,26 @@ class AutoencoderWrapper(WrapperBase):
...
-    ae_width = int(math.ceil(width / self.autoencoder_factor))
-    ae_channels = 24  # TODO(piotrmilos): make it better
-    observ_shape = (batch_size, ae_height, ae_width, ae_channels)
     self._observ = self._observ = tf.Variable(
-        tf.zeros(observ_shape, tf.float32), trainable=False)
+        tf.zeros((len(self),) + self.observ_shape, self.observ_dtype),
+        trainable=False)
     with tf.variable_scope(tf.get_variable_scope(), reuse=tf.AUTO_REUSE):
       autoencoder_hparams = autoencoders.autoencoder_discrete_pong()
+      trainer_lib.add_problem_hparams(autoencoder_hparams,
```
lukaszkaiser: Let's postpone these changes to another CL, there's enough other ones for now.

tensor2tensor/rl/trainer_model_based_test.py
```diff
@@ -17,6 +17,9 @@
 from __future__ import division
 from __future__ import print_function
 
+import os
```
lukaszkaiser: Please do not change this file in this PR.

tensor2tensor/rl/envs/tf_atari_wrappers.py
koz4k: Of course. This import is needed just for the `add_problem_hparams` function, which sets `problem` and `problem_hparams` in model `hparams`. I used this function to avoid code repetition, but I can instead manually set those things to avoid this dependency.

tensor2tensor/rl/envs/batch_env.py
koz4k: We're using the whole `observation_space` in InGraphBatchEnv, as it's passed do `utils.parse_{dtype,shape}` to determine observation datatype and shape.

PR #1024, opened 2018-08-28
Create tensor2tensor/models/video/__init__.py

cclauss:
This file is required for __tensor2tensor.models.video__ to be a valid Python module.

@lukaszkaiser Was this a step missing from https://github.com/tensorflow/tensor2tensor/commit/03e889baf68664dda80414fcefd0d8bbaa5355c3 ?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -47,7 +47,7 @@ install:
   - pip install "gym[atari]"
   # Make sure we have the latest version of numpy - avoid problems we were
   # seeing with Python 3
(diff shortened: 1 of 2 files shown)


afrozenator:
@rsepassi just fixed the missing __init__.py as well -- https://github.com/tensorflow/tensor2tensor/commit/5874f6623f12455ade2680ef02d77a8658af6ea8

Thanks for spotting it too, @cclauss 

We are hesitant about adding the matplotlib dependency

PR #1035, opened 2018-09-02
small refactor in speech problems

tlatkowski:
Small refactor in speech models.

1. Moved AudioEncoder to the separate script audio_encoder (similar to text_encoders).

2. Added tqdm for minitoring input audio files preprocessing in Librispeech problem.


afrozenator:
Hi @tlatkowski -- We're reverting the tqdm changes currently since this isn't available internally at the moment. We'll take the rest of the changes.

Thanks!

PR #1043, opened 2018-09-06
Adding short ppo training on real env between epochs

piotrmilos:
Adding short ppo training on real env between epochs


piotrmilos:
I signed it!

PR #1049, opened 2018-09-08
Updating new_problem.md to correct file names and simpler wording

bwindsor22:
Just a minor change! Main purpose is to remove a reference to `word2def`, a file that's not actually in the tutorial. The rest is just small ease-of-reading fixes: I adjust a typo before the code snippet, and match tenses after the snippet.


bwindsor22:
I signed it!

afrozenator:
Thanks @bwindsor22 !

PR #1055, opened 2018-09-10
Avoid NaN while adding sinusoidal timing signals

peakji:
Tensors with less than 4 `channels` will get **NaN** in positional encodings.


afrozenator:
Thanks @peakji !

PR #1063, opened 2018-09-12
Allow specifying which protocol to be used in tf.train.Server

changlan:
The patch allows user to specify which protocol to be used for distributed training. This enables the use of custom communication protocols (e.g. VERBS)

### DIFF
--- a/tensor2tensor/utils/trainer_lib.py
+++ b/tensor2tensor/utils/trainer_lib.py
@@ -391,7 +391,8 @@ class T2TExperiment(object):
     server = tf.train.Server(
         config.cluster_spec,
         job_name=config.task_type,
-        task_index=config.task_id)
+        task_index=config.task_id,
+        protocol=self._hparams.protocol)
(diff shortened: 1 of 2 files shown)


changlan:
@afrozenator Done!

afrozenator:
Thanks for the PR @changlan !

tensor2tensor/utils/trainer_lib.py
```diff
@@ -391,7 +391,8 @@ def run_std_server(self):
     server = tf.train.Server(
         config.cluster_spec,
         job_name=config.task_type,
-        task_index=config.task_id)
+        task_index=config.task_id,
+        protocol=self._hparams.protocol)
```
jinliangwei: This change seems to breaks with commit [730201a7d28e2ded76145ae7dd7a663311c86d4a](730201a7d28e2ded76145ae7dd7a663311c86d4a).  `self._hparams` is `None`.

PR #1067, opened 2018-09-14
Fix world model evaluation

koz4k:
Now the world model is evaluated by comparing frames from the simulated environment and those from the real environment, read from disk.
The evaluation trajectory starts from a random frame.
When resetting SimulatedBatchEnv, a number of extra frames after the initial ones are read and used as the reference trajectory in world model evaluation (class InitialFrameChooser).
Actions played in the simulated environment are taken from real trajectories, this is done by mocking the policy network to output fixed actions.


lukaszkaiser:
Merged through other PRs, thanks!

Issue #1068, opened 2018-09-16
tensor2tensor CNN_Dailymail32k summarization bug


Littlenova:
### Description
The data generation for the CNN_Dailymail32k problem fails due to a unicode related error when writing to file.

...

### Environment information
$ nvidia-smi (within docker container)
Sun Sep 16 10:32:31 2018       
+-----------------------------------------------------------------------------+
| NVIDIA-SMI 390.48                 Driver Version: 390.48                    |
|-------------------------------+----------------------+----------------------+
| GPU  Name        Persistence-M| Bus-Id        Disp.A | Volatile Uncorr. ECC |
| Fan  Temp  Perf  Pwr:Usage/Cap|         Memory-Usage | GPU-Util  Compute M. |
|===============================+======================+======================|
|   0  GeForce GTX 1050    Off  | 00000000:01:00.0  On |                  N/A |
| 30%   36C    P8    N/A /  75W |    383MiB /  1999MiB |      0%      Default |
+-------------------------------+----------------------+----------------------+
|   1  GeForce GTX 750 Ti  Off  | 00000000:03:00.0 Off |                  N/A |
| 29%   32C    P8     1W /  38W |      1MiB /  2002MiB |      0%      Default |
+-------------------------------+----------------------+----------------------+
                                                                               
+-----------------------------------------------------------------------------+
| Processes:                                                       GPU Memory |
|  GPU       PID   Type   Process name                             Usage      |
|=============================================================================|
+-----------------------------------------------------------------------------+

$ uname -a
Linux <username> 4.15.0-33-generic #36-Ubuntu SMP Wed Aug 15 16:00:05 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux

$ nvidia-docker -v
Docker version 18.06.1-ce, build e68fc7a

```
OS: Ubuntu 18.04.1 LTS
Memory: 15.6 GiB (DDR4)
Processor: Intel® Core™ i5-8500 CPU @ 3.00GHz × 6
Graphics: GeForce GTX 1050/PCIe/SSE2
GNOME: 3.28.2
OS type: 64-bit
Disk: 471.5 GB

$pip freeze | grep tensor
tensor2tensor==1.9.0
tensorboard==1.10.0
tensorflow-gpu==1.10.1


$ python -V
Python 2.7.12
```

### For bugs: reproduction and error logs

```
# Steps to reproduce:
...
Create a docker container with: (09/15/18)
[...]

PR #1086, opened 2018-09-20
Avoid a ascii codec error in CNN/DailyMail.

shahzeb1:
When a basic `t2t-trainer` command is run with the `--problem=summarize_cnn_dailymail32k` command, the following error occurs:

```
UnicodeEncodeError: 'ascii' codec can't encode character '\xbb' in position 1426: ordinal not in range(128)
```

This minor fix resolves this issue.

Fixes issue #1068


afrozenator:
Thanks @shahzeb1 !

PR #1098, opened 2018-09-26
Remove a deadlock during saving debug video

koz4k:
The deadlock was due to the output pipe buffer from FFMPEG filling up, preventing the process from consuming next input frames.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,20 +13,20 @@ env:
     - T2T_PROBLEM=algorithmic_reverse_binary40_test
     - T2T_DATA_DIR=/tmp/t2t-data
     - T2T_TRAIN_DIR=/tmp/t2t-train
-    - TF_LATEST="1.10.*"
+    - TF_LATEST="1.11.*"
     # This is necessary to have gsutil work with Python 2.7
     - BOTO_CONFIG=/dev/null
   matrix:
     # We test against recent versions of TensorFlow and tf-nightly.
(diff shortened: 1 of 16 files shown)


afrozenator:
Thanks @koz4k !

afrozenator:
I presumed the merge went through and submitted internally as well, resulting in the internal version going before this one was merged by  GitHub -- Sorry about that.

To see a list of changes please see: https://github.com/tensorflow/tensor2tensor/commit/4c8216d568abf0e8bbc198b149a9840aee19fce9

mbz:
Hi @koz4k @piotrmilos @lukaszkaiser

This PR seems to break the order of the frames i.e. the saved output video has shuffled frames (the video is extremely jumpy). if the generation speed is slower than decoding it might be fine, otherwise the order that the threads return the encoded frames seems to be random. We (@afrozenator  and I) confirmed this by reverting this change and seeing a very smooth video.

Let us revert the change and properly fix this. We also couldn't figure out the deadlock issue -- can you try to maintain the order of the threads? (The fifo commands don't seem to be doing that).

koz4k:
Hi @mbz,

Thank you for reporting this. Do you have some code snippet that reproduces the bug?

mbz:
Upon more investigation it turned out to be a False alarm. Sorry and thanks :)

koz4k:
I'm glad to hear that, thank you :)

PR #1104, opened 2018-10-01
example of invocation for ppo

piotrmilos:


tensor2tensor/rl/trainer_model_free.py
```diff
@@ -32,11 +43,38 @@
...
+    "environment_eval_spec",
+    gym_problems.standard_atari_env_eval_spec("PongNoFrameskip-v4"))
+
+  return hparams
 
 def main(_):
   hparams = trainer_lib.create_hparams(FLAGS.hparams_set, FLAGS.hparams)
-  rl_trainer_lib.train(hparams, FLAGS.problem, FLAGS.output_dir)
-
+  rl_trainer_lib.train(hparams, FLAGS.output_dir, FLAGS.output_dir)
```
afrozenator: line accidentally changed?

piotrmilos: It is a bug fix.

afrozenator: Ah! Just looked at the method signature now.

afrozenator:
Thanks @piotrmilos !

PR #1111, opened 2018-10-03
Speed up autoencoder test

konradczechowski:
Speed up autoencoder test (in model-based rl pipeline). Enable passing autoencoder hparams by flags; dynamically construct autoencoder related problems; autoencoder wrappers are constructed with hparams_set.


afrozenator:
Thanks @konradczechowski !

PR #1123, opened 2018-10-09
Add reward clipping and observation preprocessing in T2TGymEnv

konradczechowski:
Does not move preprocesing in model-based rl to T2TGymEnv, jet. 

Note that PIL.Image.resize(..., resample=BOX), should be identical to tf.image.resize(..., method=AREA) used so far, when image dimensions are dividable by resize factor (for Atari this is eg. 2,4,5), but will slightly differ for other values.


tensor2tensor/data_generators/gym_env.py
```diff
@@ -24,6 +24,8 @@
 import random
 
 import numpy as np
+from PIL import Image
```
lukaszkaiser: Let's try to do this in TF in a similar way to image encoding. I actually don't have a preference here, but there are many resizing methods and many implementation differences. We know our current tests run well with tf.resize with method=AREA so I'd rather keep it the same -- the less we change, the less potential causes if it doesn't work after refactor.

tensor2tensor/data_generators/gym_env.py
```diff
@@ -315,30 +317,58 @@ class T2TGymEnv(T2TEnv):
...
     self._envs = envs
 
-    self.observation_space = envs[0].observation_space
-    if not all(env.observation_space == self.observation_space for env in envs):
+    orig_observ_space = envs[0].observation_space
+    if not all(env.observation_space == orig_observ_space
+               for env in self._envs):
       raise ValueError("All environments must use the same observation space.")
 
+    self.observation_space = self._derive_observation_space(orig_observ_space)
```
lukaszkaiser: I think we also need to update the num_channels property now as it's set to 3.

tensor2tensor/data_generators/gym_env.py
```diff
@@ -315,30 +317,58 @@ class T2TGymEnv(T2TEnv):
...
+               dtype=orig_observ_space.dtype)
+
   def __str__(self):
     return "T2TGymEnv(%s)" % ", ".join([str(env) for env in self._envs])
 
   def _preprocess_observations(self, obs):
-    # TODO(lukaszkaiser): Implement.
-    return obs
+    assert obs.dtype == np.uint8, "Image.fromarray() requires np.uint8 dtype" \
+                                  " to work as expected"
```
lukaszkaiser: While this assert may be fine in principle, I'm worried if we'll really never cast to int32 somewhere on the way? Should we rather re-cast and warn?

tensor2tensor/data_generators/gym_env.py
```diff
@@ -315,30 +317,58 @@ class T2TGymEnv(T2TEnv):
...
+      height, width = self.observation_space.shape[:2]
+      im = im.resize(size=(width, height), resample=Image.BOX)
+      if self.grayscale:
+        im = im.convert(mode='L')
+      preprocessed.append(np.array(im))
+    return np.stack(preprocessed)
 
   def _preprocess_rewards(self, rewards):
-    # TODO(lukaszkaiser): Implement.
+    if self.clip_rewards:
```
lukaszkaiser: We'll need both clipped and unclipped rewards, one for training the other for evaluation. I suggest we store both, also on disk, and include a "reward_unclipped" field in the Frame enum.

konradczechowski:
Closing this PR to create new one (with branch rebased to master)

PR #1158, opened 2018-10-19
Passing hook context to train and eval hooks

stefan-falk:
This might make my previously proposed PR https://github.com/tensorflow/tensor2tensor/pull/1154 obsolete.

In `trainer_lib.py`, the `train_hooks()` and `eval_hooks()` of the `T2TModel` are getting called. These are `static` methods which means we cannot access the `self` reference and further not the model's `hparams`. Therefore, any hook created in `train_hooks()` or `eval_hooks()` is lacking context information. At least until now I was not able to find an obvious way to retrieve additional required information for my hooks. For more context, `train_hooks()` and `eval_hooks()` should provide a `hook_context` object. 

The (arguably) simplest way at this point would be to pass a `hook_context` in `trainer_lib.py`. So instead of

```python
train_hooks += t2t_model.T2TModel.get_train_hooks(model_name)
eval_hooks += t2t_model.T2TModel.get_eval_hooks(model_name)
```
we have:
```python
hook_context = HookContextArgs(estimator=estimator, problem=problem, hparams=hparams)

train_hooks += t2t_model.T2TModel.get_train_hooks(model_name, hook_context)
eval_hooks += t2t_model.T2TModel.get_eval_hooks(model_name, hook_context)
```

where `hook_context` is

```python
class HookContext(collections.namedtuple(
  "HookContext",
  ["estimator", "problem", "hparams"])):
  pass
```

and both hook-providing methods are getting the same, which would allow the following:

```python
import tensorflow as tf
from tensor2tensor.utils import registry
from tensor2tensor import models


class HookWithContext(tf.train.SessionRunHook):

  def __init__(self, hook_context):
    self.estimator = hook_context.estimator
    self.hparams = hook_context.hparams

  def begin(self):
    p = self.hparams
    print(p.model_dir)  # Model location
    print(p.problem.filepattern(p.data_dir, 'test'))  # Filepattern for test set


@registry.register_model
class ShakeShakeExt(models.shake_shake.ShakeShake):

  def __init__(self, *args, **kwargs):
    super().__init__(*args, **kwargs)

  @staticmethod
  def eval_hooks(hook_context):
    return [ HookWithContext(hook_context) ]
```

### DIFF
--- a/tensor2tensor/utils/t2t_model.py
+++ b/tensor2tensor/utils/t2t_model.py
@@ -159,11 +159,11 @@ class T2TModel(base.Layer):
   # Replace the two methods below in order to add custom SessionRunHooks to
   # the training procedure.
   @staticmethod
-  def train_hooks():
+  def train_hooks(hook_context):
     return []
 
   @staticmethod
-  def eval_hooks():
+  def eval_hooks(hook_context):
     return []
 
   @property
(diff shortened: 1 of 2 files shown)


lukaszkaiser:
Great thanks for solving this so nicely!

PR #1171, opened 2018-10-25
variable_scope wrapper for avg_checkpoints

Mehrad0711:
During training, you might need to run avg_checkpoints a couple of times and generate different checkpoints. In that case, you need to reuse the variables you've made in the previous session calls.


Mehrad0711:
I signed it!

PR #1179, opened 2018-10-29
Pass data_dir to feature_encoders

stefan-falk:
The reason is that we cannot do this at the moment:

```python
class LibrispeechMod(Librispeech):

  def feature_encoders(self, data_dir):
    return {
      # ..
      'targets': MyCustomEncoder(data_dir)
    }
```

because at the moment `feature_encoders()` is getting called by passing `None` (see [Librispeech](https://github.com/tensorflow/tensor2tensor/blob/v1.9.0/tensor2tensor/data_generators/librispeech.py#L140) and [CommonVoice](https://github.com/tensorflow/tensor2tensor/blob/v1.9.0/tensor2tensor/data_generators/common_voice.py#L136))

```python
def generator_(self, data_dir, tmp_dir, datasets, eos_list=None, start_from=0, how_many=0):
    # ..
    encoders = self.feature_encoders(None)
    audio_encoder = encoders["waveforms"]
    text_encoder = encoders["targets"]
    # ..
```

### DIFF
--- a/tensor2tensor/data_generators/common_voice.py
+++ b/tensor2tensor/data_generators/common_voice.py
@@ -134,7 +134,7 @@ class CommonVoice(speech_recognition.SpeechRecognitionProblem):
 
     data_dir = os.path.join(tmp_dir, "cv_corpus_v1")
     data_tuples = _collect_data(data_dir)
-    encoders = self.feature_encoders(None)
+    encoders = self.feature_encoders(data_dir)
     audio_encoder = encoders["waveforms"]
     text_encoder = encoders["targets"]
     for dataset in datasets:
--- a/tensor2tensor/data_generators/librispeech.py
+++ b/tensor2tensor/data_generators/librispeech.py
@@ -138,7 +138,7 @@ class Librispeech(speech_recognition.SpeechRecognitionProblem):
       data_files = _collect_data(data_dir, "flac", "txt")
       data_pairs = data_files.values()
 
-      encoders = self.feature_encoders(None)
+      encoders = self.feature_encoders(data_dir)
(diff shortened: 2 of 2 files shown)


stefan-falk:
Here is how one can work around it for now

```python
from abc import abstractmethod


class SpeechRecognitionProblem(object):

    def feature_encoders(self, data_dir):
        raise Exception('This should not get called!')

    @abstractmethod
    def generate_data(self, data_dir):
        raise NotImplementedError()


class Librispeech(SpeechRecognitionProblem):

    def generate_data(self, data_dir):
        d = self.feature_encoders(None)
        print('I am %s and this is data_dir: %s' % (self.__class__.__name__, d['data_dir']))
        print('Value: %s' % str(d))


class SpeechRecognitionProblemFix(SpeechRecognitionProblem):

    def __init__(self):
        self.data_dir = None

    def feature_encoders(self, data_dir):
        return {'data_dir': self.data_dir}


class LibrispeechMod(SpeechRecognitionProblemFix, Librispeech):

    def generate_data(self, data_dir):
        self.data_dir = data_dir
        Librispeech.generate_data(self, data_dir)


modified = LibrispeechMod()
modified.generate_data('/data/tensor2tensor/problems/librispeech/data')
```

Output:

```
I am LibrispeechMod and this is data_dir: /data/tensor2tensor/problems/librispeech/data
Value: {'data_dir': '/data/tensor2tensor/problems/librispeech/data'}
```

afrozenator:
Thanks @stefan-falk !

stefan-falk:
@afrozenator You're welcome. Thanks for accepting the PR :)

PR #1182, opened 2018-10-31
Return ed.RandomVariables in initializers and operate on them in regu…

dustinvtran:
…larizers.

@markvdw @dusenberrymw @danijar @balajiln: PTAL. After approval, I will merge the PR via an internal process.

@afrozenator: I'm sending this as a GitHub pull request for external collaborators to review and add comments. I'm not sure this change works with Travis, as the test now requires tfp as an import. Has that Q been resolved?

### DIFF
--- a/tensor2tensor/layers/bayes.py
+++ b/tensor2tensor/layers/bayes.py
@@ -19,7 +19,10 @@ from __future__ import absolute_import
 from __future__ import division
 from __future__ import print_function
 
+import functools
 import tensorflow as tf
+
+from tensorflow_probability import edward2 as ed
 
 
 class Softplus(tf.keras.constraints.Constraint):
@@ -110,11 +113,7 @@ class TrainableNormal(TrainableInitializer):
       raise ValueError('A TrainableInitializer must be built by a layer before '
                        'usage, and is currently only compatible with Bayesian '
                        'layers.')
-    noise = tf.random_normal(self.shape, dtype=self.dtype, seed=self.seed)
-    output = self.mean + self.stddev * noise
-    # TODO(trandustin): Hack to store parameters so KL reg. can operate on them.
-    output._parameters = (self.mean, self.stddev)  # pylint: disable=protected-access
-    return output
+    return ed.Normal(loc=self.mean, scale=self.stddev)
 
   def get_config(self):
     return {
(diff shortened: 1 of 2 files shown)


tensor2tensor/layers/bayes.py
```diff
@@ -98,11 +105,7 @@ def build(self, shape, dtype=None, add_variable_fn=None):
         trainable=True)
 
   def __call__(self):
-    noise = tf.random_normal(self.shape, dtype=self.dtype, seed=self.seed)
-    output = self.mean + self.stddev * noise
-    # TODO(trandustin): Hack to store parameters so KL reg. can operate on them.
-    output._parameters = (self.mean, self.stddev)  # pylint: disable=protected-access
-    return output
+    return ed.Normal(loc=self.mean, scale=self.stddev)
```
dusenberrymw: I was thinking that we could create and store this in the `build` function to remove a bit of Python overhead in eager mode, but it looks like Edward2 RVs only sample from the underlying distribution on the first access in eager mode due to [`rv.value` using a memoization strategy by storing the sample on the first access](https://github.com/tensorflow/probability/blob/master/tensorflow_probability/python/edward2/random_variable.py#L162).  It seems like it could potentially simplify some use cases if `rv.value` called `rv.distribution.sample` on each access to `rv.value`.  That being said, you definitely thought much longer on the design of Edward[1|2], so I bet there is a reason that would make this undesirable.

dustinvtran: We mostly memoize the Tensor for usability reasons, where you operate on RandomVariables as if it were any other Tensor. For example, if you call `x = ed.Normal(...)`, it may be less intuitive if `y = x + 1` and `z = x + 1` leads to different results.

On caching: There's an issue that actually arises before Ed2, which is that during `__init__`, Distributions call TF ops on its parameters. This breaks gradients if `Distribution.__init__` isn't recorded in the gradient tape. @vdumoulin raised a point about this in Brain Chat.

tensor2tensor/layers/bayes.py
```diff
@@ -140,12 +143,11 @@ def __init__(self, mean=0., stddev=1.):
...
-    mean, stddev = x._parameters  # pylint: disable=protected-access
-    variance2 = tf.square(self.stddev)
-    variance_ratio = tf.square(stddev) / variance2
-    regularization = tf.square(mean - self.mean) / (2. * variance2)
-    regularization += (variance_ratio - 1. - tf.log(variance_ratio)) / 2.
-    return regularization
+    """Computes regularization given an ed.Normal random variable as input."""
+    if not isinstance(x, ed.RandomVariable):
+      raise ValueError('Input must be an ed.RandomVariable.')
+    random_variable = ed.Normal(loc=self.mean, scale=self.stddev)
```
danijar: Any reason to not create this object in `__ini__` already instead of saving `self.mean` and `save.stddev` separately?

dustinvtran: Breaks Eager unfortunately if `self.mean` and `self.stddev` depend on trainable variables. See the above comment about caching. It would be nice if Distribution `__init__`s don't add TF ops to the graph.

PR #1194, opened 2018-11-01
Fixes UT convergence problem  (issue discussed in #1191)

MostafaDehghani:
Set the default for "use_memory_as_final_state" flag to False.
Removed the undefined/reduntant flag "use_memory_as_last_state".

NOTE: PR #1192 submitted by @rllin-fathom is basically addressing the same issue, but here I also added the condition of using memory as the final state only when the transition function is an LSTM (to avoid unintentional use of memory for other transition functions).

### DIFF
--- a/tensor2tensor/models/research/universal_transformer.py
+++ b/tensor2tensor/models/research/universal_transformer.py
@@ -416,7 +416,7 @@ def update_hparams_for_universal_transformer(hparams):
   # LSTM forget bias for lstm style recurrence.
   hparams.add_hparam("lstm_forget_bias", 1.0)
   # Uses the memory at the last step as the final output, if true.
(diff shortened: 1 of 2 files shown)


MostafaDehghani:
tl;dr
It was a just a bug and the performance of basic UT should be independent of this parameter.

```use_memory_as_final_state``` is used when we use an LSTM as the transition function (```use_memory_as_final_state=True```  means we use the memory instead of the output of LSTM as the output of the final step). I, unfortunately, set it to True in another PR and it messed up the code for the basic universal transformer. Sorry for the inconveniences!

afrozenator:
@MostafaDehghani - thanks for updating the issues! Merging this now.

PR #1196, opened 2018-11-01
Fixing weight initialization in symbols modality

mikeymezher:
Changed reuse val from true to tf.AUTO_REUSE in top to allow for proper weight initialization


afrozenator:
Thanks @mikeymezher !

PR #1231, opened 2018-11-16
 Exposing batch_shuffle_size as hparam

stefan-falk:
`batch_shuffle_size` was introduced in v1.10 (https://github.com/tensorflow/tensor2tensor/blob/v1.10.0/tensor2tensor/data_generators/problem.py#L805) but we are not able to control it in any way. 

With the following changes it is possible to set `batch_shuffle_size=0` in order to disable it, or set it to an arbitrary size.


stefan-falk:
I've added another commit which checks `d_input` in decode.py. As we see in `speech_recognition.py` this may be `None`:

```
  def feature_encoders(self, _):
    return {
        "inputs": None,  # Put None to make sure that the logic in
                         # decoding.py doesn't try to convert the floats
                         # into text...
        "waveforms": audio_encoder.AudioEncoder(),
        "targets": ByteTextEncoderWithEos(),
    }
```

afrozenator:
@stefan-falk - Can you separate these out in two PRs? I would prefer a PR does logically one thing only.

afrozenator:
On second thoughts the other change is small enough, so let me take them as is -- thanks for the PR!

stefan-falk:
@afrozenator Cool :) Thank you for accepting!

PR #1235, opened 2018-11-19
Fix bAbi data generator and readme

artitw:
1. Fixed bAbi data generator so it can now download data. 
2. Added instructions for bAbi data generation in readme.


afrozenator:
Thanks a lot @artitw !

PR #1267, opened 2018-12-04
Fix a bug in document

deepbrainwaves:
It gives troubles to beginners like me. After checked code in t2t_model.py, I find out this description about function returns is wrong.


lukaszkaiser:
Thank you very much, things like that help a ton to not confuse everyone!

PR #1273, opened 2018-12-04
Fix problem_0_steps in checkpoint averages

senarvi:
Currently `t2t-avg-all` creates an average of the `train_stats/problem_0_steps` variable, of type float32, which prevents restoring variables from the averaged checkpoints. This fix will create a dummy `train_stats/problem_0_steps` variable, of type int32.


afrozenator:
Thanks @senarvi !

PR #1284, opened 2018-12-07
Custom eval hooks

theorm:
This PR lets custom evaluation stage hooks to be defined at the problem level.


lukaszkaiser:
Thank you very much for this contribution, looks like it'll be very useful!

theorm:
Thanks guys!

PR #1285, opened 2018-12-07
Implement trainer restarter and add separate eval timestep limit

koz4k:


tensor2tensor/rl/restarter.py
```diff
@@ -0,0 +1,124 @@
...
+    try:
+      with open(self._counter_path, "r") as f:
+        return tuple(
+            int(counter) for counter in f.read().split(" ")
+        )
+    except FileNotFoundError:
+      return (0, -1)
+
+  def _write_counters(self, local_step, global_step):
+    with open(self._counter_path, "w") as f:
```
lukaszkaiser: Please access all files with tf file functions, so tf.gfile.Open(...) -- -otherwise it fails on other file systems.

PR #1295, opened 2018-12-12
Add Fast Decoding  Support For Relative Dot Product

aeloyq:
Add Fast Decoding  Support For And Only For Relative Dot Product by enabling caching mechanism.
This pr have been tested on "En-Zh" case and got the same results as expected.

### DIFF
--- a/tensor2tensor/layers/common_attention.py
+++ b/tensor2tensor/layers/common_attention.py
@@ -1477,11 +1477,14 @@ def dot_product_attention(q,
     return tf.matmul(weights, v)
 
 
-def _generate_relative_positions_matrix(length, max_relative_position):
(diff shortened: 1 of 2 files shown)


aeloyq:
Check CLA again.

afrozenator:
thanks a lot @aeloyq !

aeloyq:
There's a typo in 

> tensor2tensor/models/transformer.py line-276


`if self._hparams.self_attention_type not in ["dot_product", "dot_product_relative1"]:`

should be modified to:

`if self._hparams.self_attention_type not in ["dot_product", "dot_product_relative"]:`

afrozenator:
Thanks - yes @lukaszkaiser pointed that out, we fixed it on our end.

PR #1307, opened 2018-12-18
Fixing decode for extra output frames.

mbz:
Just fixed the dimension.


tensor2tensor/models/video/base_vae.py
```diff
@@ -105,7 +105,7 @@ def construct_latent_tower(self, images, time_axis):
     latent_num_frames = self.hparams.latent_num_frames
     tf.logging.info("Creating latent tower with %d frames." % latent_num_frames)
     if latent_num_frames > 0:
-      images = images[:latent_num_frames]
+      images = images[:, :latent_num_frames]
```
MechCoder: thanks for fixing! Does it make sense to add a small test for this case since this codepath is usually not used?

PR #1321, opened 2018-12-21
Add to_unicode_utf8() to text_encoder.py

cclauss:
__tensor2tensor.data_generators__ currently has two _different_ versions of __to_unicode()__, one in [__text_encoder.py__](https://github.com/tensorflow/tensor2tensor/blob/master/tensor2tensor/data_generators/text_encoder.py#L90-L94) and the other in [__wiki_revision_utils.py__](https://github.com/tensorflow/tensor2tensor/blob/master/tensor2tensor/data_generators/wiki_revision_utils.py#L38-L39).  This PR proposes renaming the second version --> __to_unicode_utf8()__ and moving it into __text_encoder.py__ with the other Unicode utility functions.

### DIFF
--- a/tensor2tensor/data_generators/cnn_dailymail.py
+++ b/tensor2tensor/data_generators/cnn_dailymail.py
@@ -24,7 +24,6 @@ import io
 import os
 import random
 import tarfile
-import six
 from tensor2tensor.data_generators import generator_utils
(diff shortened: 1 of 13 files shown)


afrozenator:
thanks a lot @cclauss for caring to clean up the code -- this is much much appreciated!

PR #1322, opened 2018-12-21
Fix bug: "is_generate_per_split" should be set to property.

kngxscn:
`is_generate_per_split` in TranslateProblem inherits from Text2TextProblem, so it should be set to property. Otherwise when we train model of TranslateProblem, the judgement in https://github.com/tensorflow/tensor2tensor/blob/65f66e70a1b145398e2df3c35f5af5fb135c7467/tensor2tensor/data_generators/text_problems.py#L302 will always be true (self.is_generate_per_split returns a method object), even if we overwrite `is_generate_per_split` (without `@property`) to return `False` in the subclass of TranslateProblem.

### DIFF
--- a/tensor2tensor/data_generators/translate.py
+++ b/tensor2tensor/data_generators/translate.py
@@ -36,6 +36,7 @@ FLAGS = tf.flags.FLAGS
 class TranslateProblem(text_problems.Text2TextProblem):
   """Base class for translation problems."""
 
+  @property
   def is_generate_per_split(self):
     return True
 


kngxscn:
I signed it!

Kotober:
Hi, I faced the same error of _ImportError: No module named discrete_domains_, when I did PR.

Don't you know how to fix it?

kngxscn:
> Hi, I faced the same error of _ImportError: No module named discrete_domains_, when I did PR.
> 
> Don't you know how to fix it?

Sorry, I don't know, and I only see this #1319.

lukaszkaiser:
Great thanks for correcting this hard-to-see bug that probably affected many people!

kngxscn:
It's my pleasure!

PR #1331, opened 2018-12-29
In the models/transformer.py, a model called Transformer Encoder was created .

Kotober:
A new class called Transformer Regressor was created in the models/transformer.py in order to conduct a regression method such as sentiment analysis. 
This is done by inheriting from Transformer Encoder and this is supposed to produce a real number as an output tensor with the shape of (N, 1,1,1) from input tensor.


Kotober:
I signed it!

PR #1345, opened 2019-01-06
Reduce usage of tf.contrib

lgeiger:
`tensor2tensor` uses many deprecated functions from `tf.contrib`.
This PR replaces the use of `tf.contrib` with stable tensorflow functionality.


afrozenator:
This is really awesome, thanks again @lgeiger !

PR #1347, opened 2019-01-06
Fix dopamine import

lgeiger:
This *should* make Travis green again.

### DIFF
--- a/tensor2tensor/rl/dopamine_connector.py
+++ b/tensor2tensor/rl/dopamine_connector.py
@@ -22,7 +22,7 @@ from __future__ import print_function
 import copy
 
 from dopamine.agents.dqn import dqn_agent
-from dopamine.discrete_domains import run_experiment
+from dopamine.atari import run_experiment
 from dopamine.replay_memory import circular_replay_buffer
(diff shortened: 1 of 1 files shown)


lgeiger:
> This should make Travis green again.

Unfortunately it doesn't :(
But the fix should still be valid.

afrozenator:
Thanks for pointing me out to this!

The correct fix is actually tricky -- since internally Dopamine is at 2.0 but they haven't pip released that version yet. I'm trying to come up with a fix that would work both internally and externally.

I'll close this once my fix works -- thanks again @lgeiger

lgeiger:
> The correct fix is actually tricky -- since internally Dopamine is at 2.0 but they haven't pip released that version yet. I'm trying to come up with a fix that would work both internally and externally.

Thanks for the clarification :+1:

afrozenator:
I pushed a change, https://github.com/tensorflow/tensor2tensor/commit/7b6ef710cf5cc79cefc2b012e2b396d7c3cd782e#diff-92a539a939ff4db57bdbe453e6482544 the discrete_domains change is hidden internally.

Closing this for now, but thanks again for pointing this out @lgeiger !

PR #1350, opened 2019-01-07
Reduce usage of tf.contrib.layers

lgeiger:
This PR replaces `tf.contrib.layers` with their stable `tf.layers` and `tf.initializers` equivalents.

### DIFF
--- a/tensor2tensor/layers/common_layers.py
+++ b/tensor2tensor/layers/common_layers.py
@@ -3561,21 +3561,21 @@ def cyclegan_upsample(net, num_outputs, stride, method="conv2d_transpose"):
       net = tf.image.resize_nearest_neighbor(
           net, [stride[0] * height, stride[1] * width])
       net = tf.pad(net, spatial_pad_1, "REFLECT")
-      net = tf.contrib.layers.conv2d(
-          net, num_outputs, kernel_size=[3, 3], padding="valid")
+      net = tf.layers.conv2d(
+          net, num_outputs, (3, 3), activation=tf.nn.relu)
     elif method == "bilinear_upsample_conv":
       net = tf.image.resize_bilinear(net,
                                      [stride[0] * height, stride[1] * width])
       net = tf.pad(net, spatial_pad_1, "REFLECT")
-      net = tf.contrib.layers.conv2d(
-          net, num_outputs, kernel_size=[3, 3], padding="valid")
+      net = tf.layers.conv2d(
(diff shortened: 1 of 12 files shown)


tensor2tensor/layers/common_layers.py
```diff
@@ -3561,21 +3561,21 @@ def cyclegan_upsample(net, num_outputs, stride, method="conv2d_transpose"):
...
+          net, num_outputs, (3, 3), activation=tf.nn.relu)
     elif method == "conv2d_transpose":
       # This corrects 1 pixel offset for images with even width and height.
       # conv2d is left aligned and conv2d_transpose is right aligned for even
       # sized images (while doing "SAME" padding).
       # Note: This doesn"t reflect actual model in paper.
-      net = tf.contrib.layers.conv2d_transpose(
-          net, num_outputs, kernel_size=[3, 3], stride=stride, padding="valid")
+      net = tf.layers.conv2d_transpose(
+          net, num_outputs, (3, 3), strides=stride, activation=tf.nn.relu)
```
lgeiger: This change makes the [`CycleGANUpsampleConv2dTranspose`](https://github.com/tensorflow/tensor2tensor/blob/master/tensor2tensor/layers/common_layers_test.py#L681) test fail with [`AttributeError: 'tuple' object has no attribute 'ndims'`](https://travis-ci.org/tensorflow/tensor2tensor/jobs/476631363#L2261).

In the unit test `cyclegan_upsample` gets called with a numpy input, tensorflow propably tries to interpret it as a list, which makes the test fail. Using a tensor as input would fix the unit test.
```python
# this fails
upsampled_output = common_layers.cyclegan_upsample(
    random_input, output_filters, stride, "conv2d_transpose")

# this works
upsampled_output = common_layers.cyclegan_upsample(
    tf.convert_to_tensor(random_input), output_filters, stride, "conv2d_transpose")
```

@afrozenator Is this a bug that should be reported to the Tensorflow team, or is this expected behavior with `tf.layers`?

afrozenator: Let's just fix the test?

lgeiger: :+1: fixed it

I just wanted to bring this to your attention since the failure doesn't happen in the other upsampling tests.

afrozenator:
Thanks a lot of this and other PRs @lgeiger - much appreciated!

PR #1359, opened 2019-01-10
Model-Based RL: Player - run real and simulated env side-by-side 

konradczechowski:


konradczechowski:
I needed to create special class for simulated (not batched) gym environment, for initial_frames injection. (`SimulatedGymEnv`) @koz4k  will probably make batched version of this class in near future, and then we probably can remove single-threaded one.

PR #1360, opened 2019-01-10
Fix tf.py_func typo

lgeiger:
Currently unit tests on master fail with [`AttributeError: module 'tensorflow' has no attribute 'py_function'`](https://travis-ci.org/tensorflow/tensor2tensor/jobs/477996581#L2440).

Unfortunately this is hard to spot since Python 2 builds are failing due to #1319 and Python 3 builds are cluttered with `inspect.getargspec()` deprecation warnings from tensorflow.

### DIFF
--- a/tensor2tensor/utils/sari_hook.py
+++ b/tensor2tensor/utils/sari_hook.py
@@ -214,7 +214,7 @@ def get_sari(source_ids, prediction_ids, target_ids, max_gram_size=4):
     return (np.asarray(sari_scores), np.asarray(keep_scores),
             np.asarray(add_scores), np.asarray(deletion_scores))
 
-  sari, keep, add, deletion = tf.py_function(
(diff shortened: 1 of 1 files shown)


afrozenator:
No I don't think this is a typo, the original author wrote the code against tf-nightly / tf 1.13 -- and its release is imminent, so the best solution would just be to wait and let 1.13 release

lgeiger:
> No I don't think this is a typo, the original author wrote the code against tf-nightly / tf 1.13 -- and its release is imminent, so the best solution would just be to wait and let 1.13 release

Thanks for the explanation, that makes sense.

Just as a side note (since you're probably aware of this), it would be quite helpful for outside contributors if internal `tensor2tensor` and open-source `tensor2tensor` would share the same CI config or at least the same requirements.

PR #1361, opened 2019-01-11
Add labels to ml-engine job

lgeiger:
Cloud ML Engine can use [labels](
https://cloud.google.com/ml-engine/docs/tensorflow/resource-labels) to organize and filter jobs.

This PR adds labels for `model`, `problem` and `hparams_set`.


afrozenator:
Thanks @lgeiger !

PR #1362, opened 2019-01-11
Mixed Precision Support for Transformer

etragas-fathom:
Hi T2T team,

This is a first-pass attempt at providing mixed-precision support to Transformer on tensor2tensor.  Per discussion on gitter with @lukaszkaiser, this branch does *not* yet achieve all the speedups we expect, but he suggested we put up this branch and perhaps the t2t team will see something subtle (or obvious...) that we missed and we can all get performance where it should be, and merge strong mixed precision support into t2t.

## Our baseline:
* We have been gradually trying to replicate the results of fairseq’s “Scaling Neural Machine Translation”, https://arxiv.org/abs/1806.00187.  This paper demonstrates increased training speed (both wall clock and fundamental efficiency) on transformer translation through mixed precision and some other tricks.  Table 1 is the core result. (Trying to simply replicate their 8x V100 results pointed us to the need to enable https://github.com/tensorflow/tensor2tensor/blob/acde95f6cea575c1e5009d7a16d95545a23e0552/tensor2tensor/bin/t2t_trainer.py#L72 on 8x; with the flag, t2t is close to parity, without, it is ~½ speed of FB results.)
* In their paper, fairseq shows ~3x speedup (steps/s) from mixed precision (over full precision)
* We picked up Fairseq’s repo and did replicate their mixed precision 8x throughput results on a Docker image. 
* We have not been able to fully replicate on t2t (the branch with this PR).

## Our results:
* All below results are 1x V100, to simplify (although we did run some 8x to confirm things are not magically fixed in that scenario)
* We tried to match hparams between t2t & fairseq, including batch size, as closely as possible (see later notes on hparams)
* T2t, full precision -> mixed: ~2.2 steps/s -> ~3.3 steps/s (~1.5x speedup)
* Fairseq, full precision -> mixed: ~2.6 steps/s -> ~6.2 steps/s (~2.4x speedup)
* NOTE #1: this observed fairseq improvement grows to ~2.7, close to paper, on 8x
* NOTE #2: steps/s between t2t and fairseq were a little hard to compare apples:apples, based on fairseq reported/reports various numbers, but I believe the relative speedups are still meaningful

## Our approach: 
* We tried to follow the guidelines in https://docs.nvidia.com/deeplearning/sdk/mixed-precision-training/#tensorflow, which fairly well mirrors some of the logic t2t already has in place (https://github.com/tensorflow/tensor2tensor/blob/acde95f6cea575c1e5009d7a16d95545a23e0552/tensor2tensor/utils/t2t_model.py#L264).  It also seems to have been fairly successfully applied to the BERT repo (https://github.com/google-research/bert/pull/255, “--use-fp16 on its own gives nearly a 2.5x performance boost and it runs smoothly.”, although we did not directly review this claim).

## Diagnosing:
* We went fairly deep into t2t, and tried disabling (not included in this PR) various casts (such as in compute_grads) and other activities that could even vaguely conceivably be having a bad interaction effect.  No improvement (obviously, we may have missed something…).
* As a separate test (not included in this branch), we also tried using the alternate nvidia approach in https://nvidia.github.io/OpenSeq2Seq/html/mixed-precision.html#prerequisites, but didn’t meaningfully change throughput
* Printing out all of the variables in the custom getter shows all variables are returned as fp16.  For diagnosis purposes, this printing code is left in the custom_getter and compute_gradient functions; will be removed before merge.

## Test environment:
* Tested variously on CUDA 9.2 & 10 (since >CUDA 9.0 is needed for full mixed precision support; see https://nvidia.github.io/OpenSeq2Seq/html/mixed-precision.html#prerequisites; we also verified this the hard way…)
* TF1.12
* Ubuntu 16 (Python 3.5) & 18 (Python 3.6)
* Problem = translate_ende_wmt32k
* Hparams = transformer_fairseq_fp16_activation_big (note: this is included in branch; before merge, this could conceivably be removed)
* (note: only change here from the base hparam set is that we set mixed precision, and set the token count to match fairseq’s run, https://github.com/pytorch/fairseq/blob/b87c536651649ad71dd8766a409ef1c032b55afa/examples/translation/README.md#L164)
* Model = transformer
* Command to test with (assuming Cuda > 9.0, on Ubuntu 16 / 18 and TF1.12)
```
python3 tensor2tensor/bin/t2t-trainer --problem="translate_ende_wmt32k" --model="transformer" --hparams_set="transformer_fairseq_fp16_activation_big" --train_steps="5000" --eval_early_stopping_steps="10000" --eval_steps="3" --iterations_per_loop="2000" --keep_checkpoint_max="80" --local_eval_frequency="2000" --generate_data="False" --tmp_dir="t2t_tmp_dir" --data_dir="<data_dir>" --output_dir="output_dir”
```

## Known technical deficiencies/issues:
* No unit tests -- can add once design & functionality settle (didn’t put in place in case you have radical design changes to suggest)
* Only tested to successfully work with vanilla Transformer (and will almost certainly fail with many other models) -- happy to discuss either generalizing for it to work throughout the code base, or, for now, to add a gate s.t. if mixed precision is on for any model/hparams other than vanilla Transformer, it refuses to run (and then this can be relaxed in a separate PR/over time).
* Passing around weight types in certain places definitely feels hacky (something like being able to universally call self.activation_dtype, etc. would feel much better, like in https://github.com/tensorflow/tensor2tensor/blob/44f669058390bec03024baa04c1a33e91cc0909d/tensor2tensor/models/mtf_transformer.py#L119).  Given how the codebase is right now, not obvious to me how to do so, but open to suggestions.
* When the mixed precision getter is invoked, it *always* forces all variables to be cast down to fp16.  This simplifies the code base, but is a different behavior than the other getters in the t2t code base; it works in this specific use case, but possibly doesn’t generalize.  Open to discussion here.
* We’ve placed some our functionality at the top of common_attention.py. We’d like your feedback on a more appropriate location in the code base.
* For convenience, there are a few debug statements that are simply commented out within branch.  We will remove those before merge.

### DIFF
--- a/tensor2tensor/layers/common_attention.py
+++ b/tensor2tensor/layers/common_attention.py
@@ -37,6 +37,31 @@ import tensorflow_probability as tfp
 from tensorflow.python.framework import function
 from tensorflow.python.ops import inplace_ops
 
+def large_compatible_negative(tensor):
+  """
+  This function is necessary because the standard value for epsilon
+  in this module (-1e9) cannot be represented using tf.float16
+  """
+
+  if tensor.dtype == tf.float16:
+    return tf.float16.min
+  return -1e9
+
+def mixed_precision_is_enabled(activation_dtype=None, weight_dtype=None, hparams=None):
+  assert not (hparams and (activation_dtype or weight_dtype)), (
+              "Provide only hparams or activation_dtype and weight_dtype")
+
+  if hparams:
+    activation_dtype = hparams.activation_dtype
+    weight_dtype = hparams.weight_dtype
+
+  return activation_dtype == tf.float16 and weight_dtype == tf.float32
+
+def maybe_upcast(logits, activation_dtype=None, weight_dtype=None, hparams=None):
+  if mixed_precision_is_enabled(activation_dtype, weight_dtype, hparams):
+    return tf.cast(logits, tf.float32)
+  return logits
+
 # Struct containing the sequences ids and order on a batch (are send to the
 # expert to allow them to compute the bias mask)
 BatchInfo = collections.namedtuple("BatchInfo", "coordinates, order")
@@ -445,8 +470,7 @@ def add_timing_signal_1d(x,
   channels = common_layers.shape_list(x)[2]
   signal = get_timing_signal_1d(length, channels, min_timescale, max_timescale,
                                 start_index)
-  return x + signal
-
+  return x + common_layers.cast_like(signal, x)
 
 @expert_utils.add_name_scope()
 def get_layer_timing_signal_learned_1d(channels, layer, num_layers):
@@ -881,10 +905,13 @@ def attention_bias_same_segment(query_segment_id, memory_segment_id):
   Returns:
     a `Tensor` with shape [batch, 1, query_length, memory_length].
   """
-  ret = tf.to_float(
+  
+  ret = (tf.to_float(
       tf.not_equal(
           tf.expand_dims(query_segment_id, 2),
-          tf.expand_dims(memory_segment_id, 1))) * -1e9
+          tf.expand_dims(memory_segment_id, 1))) *
+          large_compatible_negative(memory_segment_id))
+
   return tf.expand_dims(ret, axis=1)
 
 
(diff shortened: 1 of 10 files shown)


etragas-fathom:
CLA should be good now

rllin-fathom:
I am ok with my commits being contributed to this project

Etragas:
I am ok with my commits being contributed to this project

lukaszkaiser:
While there's still way to go, it's great to have this in and working, thanks again!!

Lollipop321:
@etragas-fathom 
I want to ask the question that which version of tensor2tensor you are using?

PR #1363, opened 2019-01-11
Fix observation reading in SimulatedBatchGymEnv.

konradczechowski:


afrozenator:
Thanks @konradczechowski !

PR #1364, opened 2019-01-11
Model-Based RL: Player, correct mapping keyboard keys to actions.

konradczechowski:


afrozenator:
Thanks @konradczechowski !

PR #1370, opened 2019-01-15
Refactor Universal Transformer

artitw:
The Universal Transformer with ACT with different `act_type`s has been combined to be more readable. Additionally, the default value for `mix_with_transformer` hparams has been changed to a blank string rather than `None` to avoid errors.


tensor2tensor/models/research/universal_transformer.py
```diff
@@ -363,8 +363,8 @@ def update_hparams_for_universal_transformer(hparams):
   hparams.daisy_chain_variables = False  # Breaks multi-gpu in while loops.
 
   # If not None, mixes vanilla transformer with Universal Transformer.
```
cbockman: If you want to avoid usage of None, probably this comment needs to be updated as well?

cbockman: I.e., while technically true, sounds like you are trying to dissuade the user from using None

tensor2tensor/data_generators/babi_qa.py
```diff
@@ -113,7 +113,7 @@ def _prepare_babi_data(tmp_dir, data_dir):
                            "AppleWebKit/537.36 (KHTML, like Gecko) "
                            "Chrome/63.0.3239.132 Safari/537.36"}
   resp = requests.get(_URL, headers=headers)
-  with open(file_path, "wb") as f:
+  with open(file_path, 'wb') as f:
```
lukaszkaiser: Please revert this (default style in T2T are " quotes).

tensor2tensor/data_generators/babi_qa.py
```diff
@@ -463,7 +463,6 @@ def hparams(self, defaults, unused_model_hparams):
     if "context" in p.vocab_size:
       del p.vocab_size["context"]
 
-
```
lukaszkaiser: Let's not make irrelevant changes if not needed, let's revert this whole file.

tensor2tensor/data_generators/mathematical_language_understanding.py
```diff
@@ -52,9 +52,8 @@
 
 import tensorflow as tf
 
-
 @registry.register_problem
-class AlgorithmicMathTwoVariables(text_problems.Text2TextProblem):
+class MathematicalLanguageUnderstanding(text_problems.Text2TextProblem):
```
lukaszkaiser: Please keep the name as before, this is an algorithmic problem which makes it easier to locate in the large number of T2T datasets.

tensor2tensor/data_generators/mathematical_language_understanding.py
```diff
@@ -89,6 +88,7 @@ def generate_samples(self, data_dir, tmp_dir, dataset_split):
     Yields:
       The data examples.
     """
+
```
lukaszkaiser: This line looks irrelevant and generally in our style guide we discourage newlines just before start of a function, please revert the whole file if no other changes are there.

tensor2tensor/models/research/universal_transformer_util.py
```diff
@@ -1226,460 +1209,6 @@ def should_continue(u0, u1, halting_probability, u2, n_updates, u3):
 
   return new_state, (ponder_times, remainders)
 
-
```
lukaszkaiser: The code below are interesting and used functions, please do not remove it unless there's an equivalent.

artitw:
I will close this and revise accordingly. Thank you

PR #1373, opened 2019-01-16
Combine AUT variants into a single function and define mix_with_transformer

artitw:
Here I propose for the Universal Transformer with ACT with different `act_type`s to be combined into a single function to be more understandable. I believe that the proposed backwards-compatible change would benefit future variants.


lukaszkaiser:
Great thanks @artitw -- this makes the UT much more readable indeed. Great work, let us know if there're more and other things to correct!

PR #1383, opened 2019-01-19
changed _normalize_body_output to sum the mean of each loss when loss…

jackd:
… is a list

I believe this is the intended behaviour, but it may result in some changes to existing models where each loss returned in `body` was the same shape (in which case the `add_n` rather than `reduce_mean` will make the resulting loss `len(losses)` times larger.


afrozenator:
Thanks @jackd !

PR #1385, opened 2019-01-19
Add a way to set max_subtoken_length in translate_enzh problem.

kngxscn:


afrozenator:
Hi @kngxscn -- thanks for the PR, I'm accepting this but changing it a little -- just doing the changes in translate_enzh.py's generate_encoded_samples method should suffice -- max_subtoken_length is already set to 200 in Text2TextProblem, so the other changes are no-ops, I'll change them on my end.

Thanks again for finding and fixing this!

kngxscn:
@afrozenator Oh, I see. Thank you!

PR #1387, opened 2019-01-21
Use TensorFlow 1.12 image on FloydHub

lgeiger:
FloydHub now has [support for TensorFlow 1.12](https://docs.floydhub.com/guides/tensorflow/#tensorflow-112) which is require for T2T to work correctly.


afrozenator:
Thanks @lgeiger !

Issue #1219, opened 2018-11-13
Requested more than 0 entries, but params is empty.


echan00:
Trying to serve my Chinese to English model and am having trouble querying. I am receiving an error: 

```
(test) root@ubuntu-c-8-16gib-sfo2-01:~/T2T_Model# t2t-query-server   --server=0.0.0.0:9000   --servable_name=transformer   --problem=translate_enzh_wmt32k_rev   --data_dir=/root/T2T_Model/t2t_data   --inputs_once='Hello my name is John.'

Traceback (most recent call last):
  File "/usr/local/bin/t2t-query-server", line 17, in <module>
    tf.app.run()
  File "/usr/local/lib/python2.7/dist-packages/tensorflow/python/platform/app.py", line 125, in run
    _sys.exit(main(argv))
  File "/usr/local/bin/t2t-query-server", line 12, in main
    query.main(argv)
  File "/usr/local/lib/python2.7/dist-packages/tensor2tensor/serving/query.py", line 89, in main
    outputs = serving_utils.predict([inputs], problem, request_fn)
  File "/usr/local/lib/python2.7/dist-packages/tensor2tensor/serving/serving_utils.py", line 157, in predict
    predictions = request_fn(examples)
  File "/usr/local/lib/python2.7/dist-packages/tensor2tensor/serving/serving_utils.py", line 113, in _make_grpc_request
    response = stub.Predict(request, timeout_secs)
  File "/usr/local/lib/python2.7/dist-packages/grpc/_channel.py", line 533, in __call__
    return _end_unary_response_blocking(state, call, False, None)
  File "/usr/local/lib/python2.7/dist-packages/grpc/_channel.py", line 467, in _end_unary_response_blocking
    raise _Rendezvous(state, None, None, deadline)
grpc._channel._Rendezvous: <_Rendezvous of RPC that terminated with:
	status = StatusCode.INVALID_ARGUMENT
	details = "Requested more than 0 entries, but params is empty.  Params shape: [1,4,8,0,64]
	 [[{{node transformer/while/GatherNd_32}} = GatherNd[Tindices=DT_INT32, Tparams=DT_FLOAT, _output_shapes=[[?,8,?,?,64]], _device="/job:localhost/replica:0/task:0/device:CPU:0"](transformer/while/Reshape_65, transformer/while/stack)]]"
[...]

PR #1389, opened 2019-01-21
sovle the issue 1219. When use the translate_enzh_rev problem module, the decoding resulte is wrong.

googlehjx:
[E**rror Querying Server: Requested more than 0 entries, but params is empty**.](https://github.com/tensorflow/tensor2tensor/issues/1219) #1219
After days of work, I have solved this problem. It caused by the function serving_input_fn. The input example no need to be reversed any more. So I deleted the code.

### DIFF
--- a/tensor2tensor/data_generators/problem.py
+++ b/tensor2tensor/data_generators/problem.py
@@ -877,8 +877,7 @@ class Problem(object):
         dtype=tf.string, shape=[None], name="serialized_example")
     dataset = tf.data.Dataset.from_tensor_slices(serialized_example)
     dataset = dataset.map(self.decode_example)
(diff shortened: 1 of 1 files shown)


googlehjx:
> Sorry, I don't understand this change. In problems that reverse examples, how comes we should not reverse them? I think this change would break many datasets what work well with reversing. Could we not find out what's the specific thing about this one that fails?
the function serving_input_fn only affects serving process, with no affection about the training and others.

googlehjx:
I agree with that, merging my change to the code repository.

afrozenator:
Thanks a lot @googlehjx ! WilL merge this in shortly.

Issue #1394, opened 2019-01-21
Propose to contribute documentation on use and design of MultiProblem


cwbeitel:
The new MultiProblem class and instances are really cool!!

I'm interested in writing some documentation on MultiProblem both regarding use and design. In my case towards understanding how to extend it to support multiple input modalities (audio and video).

Checking whether this isn't something the authors would prefer to do themselves.

/cc @rsepassi @lukaszkaiser

afrozenator:
Hi Christopher - Please give it a shot!

Regards

PR #1399, opened 2019-01-22
First draft of multi-problem docs

cwbeitel:
vis. issue https://github.com/tensorflow/tensor2tensor/issues/1394


afrozenator:
Thanks a LOT for this @cwbeitel  -  Should we merge it in?

cwbeitel:
@afrozenator Ah I left the [WIP] on there because I hadn't run any of the commands there yet, finding little changes in doing so. Just started datagen running this afternoon so it might be a day before I've test run training as well. I'll push those changes and if you think it would be helpful to people right away feel free to merge it and I'll send a follow-on, otherwise we can let it sit a day.

afrozenator:
Makes sense to wait I think -- thanks again @cwbeitel  -- do ping me when you feel it is ready, thanks!

cwbeitel:
@afrozenator gtg!

afrozenator:
Many many thanks @cwbeitel !

PR #1401, opened 2019-01-22
added optimizer registry

jackd:
Added optimizer registration, similar to other registration methods. This allows users to add new optimizers in their `t2t_usr_dir` without having to manipulate tensor2tensor code directly.

EDIT: points below [resolved](https://github.com/tensorflow/tensor2tensor/pull/1401/commits/4d1b88f9842a099b2d4f02ecb8548d74bacabcd9)
Error checking number of args in registered function is a bit gross (`inspect` doesn't play nicely with `functools.partial`), but unsure of a better way of handling this.

I'm not convinced the way of handling optimizers in `tf.contrib.layers.OPTIMIZER_CLS_NAMES` is optimal, but I don't have a better way to bind loop variables to function calls.

### DIFF
--- a/tensor2tensor/utils/optimize.py
+++ b/tensor2tensor/utils/optimize.py
@@ -157,7 +157,7 @@ def adam_w(learning_rate, hparams):
       epsilon=hparams.optimizer_adam_epsilon)
 
 
-@registry.register_optimizer('adafactor')
+@registry.register_optimizer('Adafactor')
 def register_adafactor(learning_rate, hparams):
   return adafactor.adafactor_optimizer_from_hparams(hparams, learning_rate)
 
--- a/tensor2tensor/utils/optimize.py
+++ b/tensor2tensor/utils/optimize.py
@@ -18,12 +18,14 @@ from __future__ import absolute_import
 from __future__ import division
 from __future__ import print_function
 import numpy as np
+import functools
 
 from tensor2tensor.layers import common_layers
 from tensor2tensor.utils import adafactor
 from tensor2tensor.utils import mlperf_log
 from tensor2tensor.utils import multistep_optimizer
 from tensor2tensor.utils import yellowfin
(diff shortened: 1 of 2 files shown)


afrozenator:
This is really neat @jackd -- many thanks! Will merge this in now.

Just FYI @rsepassi just a few days ago wrote a generic registry https://github.com/tensorflow/tensor2tensor/commit/ee64f6f29884aa66c98130afaf7e4eb182e0ca1f we'll port this (and maybe other registries) over to that abstraction -- but many thanks again!

jackd:
@afrozenator cheers. I was considering abstracting away some of the registry details myself, glad I wasn't the only one.

I'm a little confused as to the point of the new implementation though - `_GENERIC_REGISTRIES`/`create_registry`/`_reset` don't seem be used anywhere except tests? `_reset` loops over hard-coded registers rather than all registered registries? There's also no way of registering registries with custom error checking (like those already implemented)?

`_OPTIMIZERS` register is looking a bit lonely down the bottom too (and so I'm not surprised it was overlooked in `_reset`).

In short, this isn't how I'd have refactored it and it looks like a mash up of competing ideas, but there may well be use cases this is trying to serve that I'm not appreciating. Happy to put in a separate PR, just don't want to step on anyone's toes.

rsepassi:
Definitely welcome a PR on cleaning this stuff up. Do you want to think
about how you’d design (or modify the just added one) a generic registry
system and send over a doc or just update this thread? We want something
simple and flexible. The new generic registry seemed like a good starting
point but it sounds like it’s too limiting. We can work towards something
better with your help.

jackd:
@rsepassi So I started drafting a plan but it got a bit out of hand so I just went ahead and made a [pull request](https://github.com/tensorflow/tensor2tensor/pull/1410). Happy to discuss/make changes, but probably best to continue conversation over there :)

PR #1410, opened 2019-01-25
Register refactor

jackd:
Re-write of `utils.registry` to a dict-like interface as per discussion [here](https://github.com/tensorflow/tensor2tensor/pull/1401).

Common code pulled into `dict`-like `Registry` class. Old interface remains (`registry.register_problem`, `registry.hparams` etc), though this does mean there is a fair bit of aliasing (e.g. `register_problem = problem_registry.register`).

Removed registry registry (i.e. the registry containing registries) - I don't see why it was necessary. The point of registries (as far as I can tell) is to allow `tensor2tensor` to play nicely with external code through `t2t_usr_dir`. If external code wants to add a registry that `tensor2tensor` knows nothing about, then `tensor2tensor` won't be calling it, so it doesn't need to be registered... unless I'm missing something.

Marked a couple of function names deprecated, since they are inconsistent with naming convention of other functions in the old registry version. All aliased functions could potentially be deprecated, though I don't see much point.


jackd:
Same tests failing on master (and as far as I can tell, unrelated to anything touched here).

rsepassi:
Excellent, thank you! At first glance, looks good! I’ll review later today.

jackd:
One possible change would be to remove `value_transformer` from `Registry` so that all values are functions. The old API (which sometimes returns functions, and sometimes their values) could be maintained in the wrapper functions instead, i.e.

```
attack_registry = Registry("attack")

def attack(name):
  return attack_registry[name]()
```

This would make the new API more consistent (always return a function), though I'm not much of an API designer....

If we accept registries will only ever register callables, we could also put in some more specific error checks at the top level (or perhaps a subclass).

I'm also thinking putting all registries in a single object for namespacing might be nice, similar to `metrics`, rather than having multiple top-level variables with `_registry` suffix.

I'll be at a proper desk in a few hours - happy to make changes myself then/take into account further comments.

tensor2tensor/utils/registry.py
```diff
@@ -44,103 +44,33 @@ class MyModel(T2TModel):
...
-    raise KeyError("No registry named %s. Available:\n%s" % (
-        registry_name, sorted(_GENERIC_REGISTRIES)))
-  return _GENERIC_REGISTRIES[registry_name]
+def default_name(class_or_fn):
+  """Default name for a class or fn.
 
-
-def create_registry(registry_name):
-  """Create a generic object registry.
+  This is the naming function by default for registers expecting classes or
```
rsepassi: s/registers/registries/g

and in other places where you refer to register (as a noun) it should be registry

tensor2tensor/utils/registry.py
```diff
@@ -151,372 +81,285 @@ def default_object_name(obj):
...
-    opt_fn = name
-    registration_name = default_opt_name(opt_fn)
-    return decorator(opt_fn, registration_name=registration_name)
-
-  return lambda opt_fn: decorator(opt_fn, name)
-
+class Registry(object):
+  """Dict-like class for managing registrations."""
+  def __init__(
+      self, register_name, default_key_fn=default_name, validator=None,
```
rsepassi: s/register_name/name/g

tensor2tensor/utils/registry.py
```diff
@@ -151,372 +81,285 @@ def default_object_name(obj):
...
+      validator (optional): if given, this is run before setting a given
+        (key, value) pair. Accepts (key, value) and should raise if there is a
+        problem. Overwriting existing keys is not allowed and is checked
+        separately.
+      on_set (optional): callback function accepting (key, value) pair
+        which is run after an item is successfully set.
+      value_transformer (optional): if run, `__getitem__` will return
+        value_transformer(key, registered_value).
+    """
+    self._register = {}
```
rsepassi: self._registry

tensor2tensor/utils/registry.py
```diff
@@ -151,372 +81,285 @@ def default_object_name(obj):
...
+    if key in self:
+      raise KeyError("key %s already registered in registry %s"
+                     % (key, self._name))
+    self.validate(key, value)
+    self._register[key] = value
+    callback = self._on_set
+    if callback is not None:
+      callback(key, value)
+
+  def register(self, key=None):
```
rsepassi: add method docstrings here and in any other method that's > a few lines.
at least a 1-liner.
if > 1 line docstring, make sure it has an Args: section and a Returns: section

tensor2tensor/utils/registry.py
```diff
@@ -151,372 +81,285 @@ def default_object_name(obj):
...
+    else:
+      return lambda value: decorator(value, key=key)
 
-def optimizer(name):
-  if name not in _OPTIMIZERS:
-    raise LookupError("Optimizer %s never registered. "
-                      "Available optimizers:\n %s"
-                      % (name, "\n".join(list_optimizers())))
-  return _OPTIMIZERS[name]
+  def _get(self, key):
```
rsepassi: rm method. can use `__getitem__` in the same way (e.g. `model = model_registry.__getitem__`)

tensor2tensor/utils/registry.py
```diff
@@ -151,372 +81,285 @@ def default_object_name(obj):
...
+                      display_list_by_prefix(sorted(self), 4)))
+    value = self._register[key]
+    return self._value_transformer(key, value)
 
-def list_optimizers():
-  return list(sorted(_OPTIMIZERS))
+  def __contains__(self, key):
+    return key in self._register
 
+  def __keys__(self):
```
rsepassi: change to `keys`. `__keys__` is not a special method i'm familiar with and google isn't finding anything on it either.

tensor2tensor/utils/registry.py
```diff
@@ -151,372 +81,285 @@ def default_object_name(obj):
...
-  hp = _HPARAMS[name]()
-  if hp is None:
-    raise TypeError("HParams %s is None. Make sure the registered function "
-                    "returns the HParams object." % name)
-  return hp
+  def __delitem__(self, k):
+    del self._register[k]
 
+  def pop(self, k):
+    return self._value_transformer(k, self._register.pop(k))
```
rsepassi: rm pop. don't want to encourage mutable registries (i.e. keep them add-only)

tensor2tensor/utils/registry.py
```diff
@@ -151,372 +81,285 @@ def default_object_name(obj):
...
 
-def list_hparams(prefix=None):
-  if prefix:
-    return [name for name in _HPARAMS if name.startswith(prefix)]
-  return list(_HPARAMS)
+  def get(self, key, d=None):
```
rsepassi: s/d/default

tensor2tensor/utils/registry.py
```diff
@@ -151,372 +81,285 @@ def default_object_name(obj):
...
-    """Registers & returns p_cls with registration_name or default name."""
-    p_name = registration_name or default_name(p_cls)
-    if p_name in _PROBLEMS and not tf.executing_eagerly():
-      raise LookupError("Problem %s already registered." % p_name)
+  Returns:
+    base_name: A string with the base problem name.
+    was_reversed: A boolean.
+    was_copy: A boolean.
+  """
+  # Recursively strip tags until we reach a base name.
```
rsepassi: can you update the var names in this method to be consistent? e.g. was_rev vs was_reversed

tensor2tensor/utils/registry.py
```diff
@@ -151,372 +81,285 @@ def default_object_name(obj):
...
+  if out is None:
+    raise TypeError("HParams %s is None. Make sure the registered function "
+                    "returns the HParams object" % key)
+  return out
+
+
+model_registry = Registry("models", on_set=_on_model_set)
+optimizer_registry = Registry(
+    "optimizers",
+    default_key_fn=lambda fn: misc_utils.snakecase_to_camelcase(fn.__name__),
```
rsepassi: let's actually rm this. have snakecase as the standard, no special case for optimizers.

tensor2tensor/utils/registry.py
```diff
@@ -151,372 +81,285 @@ def default_object_name(obj):
...
+    "attacks", value_transformer=_call_value)
+attack_params_registry = Registry(
+    "attack_params", value_transformer=_call_value)
+pruning_params_registry = Registry(
+    "pruning_params", value_transformer=_call_value)
+pruning_strategy_registry = Registry("pruning_strategies")
+
+# consistent version of old API
+model = model_registry._get
+list_models = lambda: sorted(model_registry)
```
rsepassi: let's rm sorted here and instead have `__iter__` return `sorted(self._registry)`. these can be `lambda: list(model_registry)`

tensor2tensor/utils/registry.py
```diff
@@ -151,372 +81,285 @@ def default_object_name(obj):
...
+
+# list_problems won't list all rev/copy combinations,
+# so the name is slightly confusing. Similarly, register_problem will raise an
+# error if attempting to register a value with a non-base key.
+# Keeping for back-compatibility
+list_problems = list_base_problems
+register_problem = register_base_problem
+
+
+def problem(problem_name, base_registry=base_problem_registry):
```
rsepassi: rm base_registry argument. can just use `base_problem_registry` in the fn body

rsepassi:
I think the value transformer is nice, so let's keep that for now.
Yeah, putting them all in one object `REGISTRIES = {}` sounds good to me.
Maybe even a `registry_registry`? :P

tensor2tensor/utils/registry.py
```diff
@@ -151,372 +81,285 @@ def default_object_name(obj):
...
-    _RANGED_HPARAMS[rhp_name] = rhp_fn
-    return rhp_fn
 
-  # Handle if decorator was used without parens
-  if callable(name):
-    rhp_fn = name
-    return decorator(rhp_fn, registration_name=default_name(rhp_fn))
+def _nargs_validator(nargs, message):
+  def f(key, value):
+    args, varargs, keywords, _ = inspect.getargspec(value)
```
rsepassi: use `inspect.getfullargspec` instead. looks like `getargspec` is deprecated

jackd:
All changes implemented except `sorted` `__iter__` comment above. It would need to be something like
```
def __iter__(self):
  return iter(sorted(self))

list_problems = lambda: list(Registries.problems)
```

Seems a bit weird to only expose an iterator to the sorted keys, rather than the sorted keys themselves. I suppose we could make `keys` return the sorted version, but I still feel it's a bit odd to force a less efficient implementation which is different to `dict`s on people just so `list_problems` etc. methods are slightly simpler. I don't fell `list_problems = lambda: sorted(Registries.problems)` is too bad. Membership test would be slower via keys, i.e. `x in Registries.problems.keys()`. If you want to make the change go ahead, I just need a bit more convincing...

I've updated optimizer naming convention to mostly snake case - the only exception being `SGD` and `RMSProp`, which have special rules (`sgd` and `rms_prop`). I've changed all instances in the code, and `registry.optimizer` still accepts legacy naming conventions (I'm thinking of all those logged `hparams=` flags).

Thanks for the feedback. For future reference: is this the standard work-flow for pull requests? Throw something up, get feedback, make tweaks? I'm very much a researcher used to working with very small groups... usually 1 (i.e. just me), or with someone I can shout to down the corridor... and while this hasn't been a painful experience by any stretch, if there's a nicer workflow then I'm only not using it out of ignorance rather than being "set in my ways".

rsepassi:
Fine with me. Thanks for these changes!

Yes, this is the standard workflow. Make a pull request, go through 1+ rounds of code review, merge. For things that are larger and/or require more discussion, we tend to write it up and/or talk synchronously (in-person, chat, or video chat). It's definitely different than small groups working on small codebases where everybody might be sitting right next to each other and everybody might have direct push access to the repo, but we've found it to be a good set of tradeoffs across scalability/friction/safety (every commit at Google goes through the same code review process).

jackd:
... did something weird happen with the commits here? I see `Merge branch 'master' into register_refactor` follwed by `merged commit xxx into tensorflow:master`... and then there's [this](https://github.com/jackd/tensor2tensor/compare/2aa437a16e...8a18032ba0) which has me thoroughly confused (I mean, it looks great, but not sure it has anything to do with me or registry changes). `registry.py` on [tensorflow/tensor2tensor](https://github.com/tensorflow/tensor2tensor/blob/master/tensor2tensor/utils/registry.py) fork doesn't have registry changes either...

Sorry for the confusion, still getting my head around git...

rsepassi:
No, not your fault at all. This is some weirdness with our system because
we’re syncing code back and forth between GitHub and Google’s internal
version control system. Another commit will come out tomorrow with your
change applied. Will update here when it’s out. Sorry about that.

rsepassi:
Now actually permanently merged with this [commit](https://github.com/tensorflow/tensor2tensor/commit/af17fc746f56bbbd553ba889d2291eb941331dcb).

PR #1411, opened 2019-01-25
Replace deprecated assert calls

lgeiger:
`self.assertEquals` --> `self.assertEqual`


lgeiger:
@afrozenator thanks for the fast responses - No problem, I'm happy to help!

PR #1416, opened 2019-01-28
Partial Mixed Precision for Universal Transformer

etragas-fathom:
Hey T2T team,

##Background:
This is an PR regarding Universal Transformer support for Mixed Precision added in https://github.com/tensorflow/tensor2tensor/pull/1362

After seeing promising speed increases when using mixed precision with Transformer (1.5x faster compared to Transformer without MP) , we were hoping to achieve the same using Universal Transformer. 

However, it seems there are a few blockers to this. We’ve attempted to resolve a few simple code issues, but are running up against further issues and would appreciate any feedback you have. 

## Changes so far:
1. We've added cast_like calls to transformer_util [here](https://github.com/tensorflow/tensor2tensor/compare/master...medicode:t2t-master-ut#diff-d23575af8cb2f9ebc777efa511b3c7dbL1445) and [here](https://github.com/tensorflow/tensor2tensor/compare/master...medicode:t2t-master-ut#diff-d23575af8cb2f9ebc777efa511b3c7dbL1482), these were necessary to avoid crashes caused by dtype differences.
2. We've slightly refactored the mixed precision init loss scale to be an hparam instead of a fixed constant, since the init loss scale's value depends on the model being used.
3. We've added fp16 enabled hparams [here](https://github.com/tensorflow/tensor2tensor/compare/master...medicode:t2t-master-ut#diff-750434e84a49123af08c4226eb4962f7R469)

## Current Issue:
After the above changes, UT **without** Mixed Precision disabled is 4x faster than with it enabled.

Details:
1. To run UT without Mixed Precision:
    `python3 /usr/src/t2t/tensor2tensor/bin/t2t-trainer --accelerator_preference="nvidia-tesla-v100" --data_dir="<data_dir>" --description="T2T_TEST_translate_ende_wmt32k" --eval_early_stopping_steps="10000" --eval_steps="3" --generate_data="False" --hparams="" --hparams_set="universal_transformer_base_old" --iterations_per_loop="2000" --keep_checkpoint_max="80" --local_eval_frequency="2000" --model="universal_transformer" --output_dir="gs://fathom-dev-210618-workspace-h5bnpfec/fp32_ut_ende_v100/2019-01-25T15-48-50.670464+00-00/t2t-trainer/output" --problem="translate_ende_wmt32k" --train_steps="3000" --worker_gpu="1"`

    **Resulting steps per second: ~4.7**

    Note: I have not included the `universal_transformer_base_old` hparams in this PR. But they are equivalent to `universal_transformer_base_old_fp16` with the ` hparams.activation_dtype = 'float16'` line removed.

2. To run UT with mixed precision change the hparams above to `universal_transformer_base_old_fp16`

    **Resulting steps per second Steps/sec: 0.9**

3. A second UT + Mixed Precision run was attempted, with `mixed_precision_optimizer_init_loss_scale` set to 1. This did not change the speed.

### DIFF
--- a/tensor2tensor/models/research/universal_transformer.py
+++ b/tensor2tensor/models/research/universal_transformer.py
@@ -466,6 +466,7 @@ def universal_transformer_big():
   hparams.filter_size = 8192
   return hparams
 
+
 @registry.register_hparams
 def universal_transformer_base_fp16():
   hparams = transformer.transformer_base()


Etragas:
I am ok with these commits being contributed

afrozenator:
Manually verified the CLA, looks good.

mehmedes:
@etragas-fathom
Was there actually a speed-up?
Wasn't your point that there was a slow-down?

etragas-fathom:
@mehmedes you are correct there is no speed up. This PR was just getting rid of the crashes that made it impossible to run UT. For what it's worth, I made this Issue to track the slowness https://github.com/tensorflow/tensor2tensor/issues/1417

PR #1424, opened 2019-02-01
Replace numerically more stable log1p and expm1

lgeiger:
This PR replaces `log(1 + x)` with `log1p(x)` and `exp(x) - 1` with `expm1(x)`. These functions are more precise if x is close to zero.


afrozenator:
Thanks a lot for these @lgeiger -- they are much appreciated, as always!

PR #1431, opened 2019-02-02
Travis: Allow failures for tf@1.12

lgeiger:
If I recall correctly Google internally develops against `tf-nightly`. This results in failing builds when testing against the latest stable version of Tensorflow (currently all builds fail).

This PR adds Python 3 tests for `tf-nightly` and allows failures when testing against Tensorflow 1.12.

I think this would make it easier to merge PRs and quickly see the relevant failures.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,11 +22,8 @@ env:
     - TF_VERSION="1.12.*"
     - TF_VERSION="tf-nightly"
 matrix:
-  exclude:
-    # We test against all versions in Python 2 but only the latest stable
-    # version in Python 3
-    - python: "3.6"
-      env: TF_VERSION="tf-nightly"
+  allow_failures:
+    - env: TF_VERSION=$TF_LATEST
 before_install:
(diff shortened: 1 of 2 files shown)


afrozenator:
@lgeiger Thanks for the PR -- but we are very wary of accepting this, if tf 1.13 doesn't get released soon then we'll want to make sure 1.12 is green. If it does, then this PR is meaningless (except the py3 and nightly -- which we would have).

Do you want to send just the py3 + nightly part of the PR?

Looping in @lukaszkaiser as well.

lgeiger:
> Thanks for the PR -- but we are very wary of accepting this, if tf 1.13 doesn't get released soon then we'll want to make sure 1.12 is green. If it does, then this PR is meaningless (except the py3 and nightly -- which we would have).

👍 Sure that makes sense, though we could always disallow failures again once 1.13 is stable.

> Do you want to send just the py3 + nightly part of the PR?

--> #1445

PR #1442, opened 2019-02-07
Fix Algorithmic Math Two Variables

artitw:
Mathematical Language Understand problem name change to Algorithmic Math Two Variables is now fully updated


lukaszkaiser:
Great thanks for taking care of this! Btw., I ran Transformer with the transformer_tpu setting on a TPU for 1M steps on this problem and it did get to 0.999 sequence-level accuracy. So longer training with larger batches can help. And off to harder problems!

artitw:
Thanks for doing on TPU and sharing results! I will specify that for my results 100k steps were used across all experiments for comparison purposes. Harder problems in the pipeline!
 
> Great thanks for taking care of this! Btw., I ran Transformer with the transformer_tpu setting on a TPU for 1M steps on this problem and it did get to 0.999 sequence-level accuracy. So longer training with larger batches can help. And off to harder problems!

PR #1446, opened 2019-02-09
Slightly simplify python loops and attr access

lgeiger:
Just some minor readability improvements that accumulated during debugging bugs in my custom model code.

### DIFF
--- a/tensor2tensor/data_generators/text_encoder.py
+++ b/tensor2tensor/data_generators/text_encoder.py
@@ -393,8 +393,8 @@ class TokenTextEncoder(TextEncoder):
       filename: Full path of the file to store the vocab to.
     """
     with tf.gfile.Open(filename, "w") as f:
-      for i in range(len(self._id_to_token)):
-        f.write(self._id_to_token[i] + "\n")
+      for token in self._id_to_token:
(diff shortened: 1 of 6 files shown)


tensor2tensor/data_generators/text_encoder.py
```diff
@@ -393,8 +393,8 @@ def store_to_file(self, filename):
       filename: Full path of the file to store the vocab to.
     """
     with tf.gfile.Open(filename, "w") as f:
-      for i in range(len(self._id_to_token)):
-        f.write(self._id_to_token[i] + "\n")
+      for token in self._id_to_token:
+        f.write(token + "\n")
```
afrozenator: this will not work, _id_to_token is a dict and the previous code basically writes tghe vocab file in order.

afrozenator:
Thanks a lot again for this @lgeiger , I'm merging in now, will do that revert on my end.

tensor2tensor/data_generators/text_encoder.py
lgeiger: Good catch. I missed the fact that dicts aren't order prior to Python 3.6.

PR #1447, opened 2019-02-12
Added log information of performance of inference of decoding, if bat…

cuixiaom:
…ch_size=1 log average latency performance, if batch_size>1, log throughput performance


lukaszkaiser:
Great thanks, this is very helpful!

PR #1451, opened 2019-02-13
Replace .append loop with list comprehension

lgeiger:
This PR replaces loops that only append to a list with list comprehensions. This makes it more concise and (arguably) more readably.

List comprehensions also are quite a bit faster than appending to a list. Toy example in Python 3.6:
```python
In [1]: %%timeit
   ...: l = []
   ...: for i in range(5000):
   ...:     l.append(i)
375 µs ± 1.73 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)

In [2]: %%timeit
   ...: l = [i for i in range(5000)]
168 µs ± 1.08 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)
```

### DIFF
--- a/tensor2tensor/data_generators/cnn_dailymail.py
+++ b/tensor2tensor/data_generators/cnn_dailymail.py
@@ -118,9 +118,7 @@ def example_splits(url_file, all_files):
 
   all_files_map = {f.split("/")[-1]: f for f in all_files}
 
-  urls = []
-  for line in tf.gfile.Open(url_file):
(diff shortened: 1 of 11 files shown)


afrozenator:
Hi @lgeiger -- Can you resolve the conflicts here? Sorry for the delay in merging.

lgeiger:
Rebased 👍

afrozenator:
Thanks @lgeiger !

PR #1454, opened 2019-02-15
Prefer literals over list and dict functions

lgeiger:
This is just some search and replace to improve readability:
- `d = dict()` --> `d = {}`
- `l = list()` --> `l = []`

Some nice benefit of this is that literals in Python are quite a bit faster (though only 60 ns - 75 ns):
```shell
In [1]: %timeit d = dict()
101 ns ± 0.0829 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)

In [2]: %timeit dl = {}
34.1 ns ± 0.378 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)

In [3]: %timeit l = list()
101 ns ± 0.541 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)

In [4]: %timeit l = []
25.4 ns ± 0.346 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)
```

### DIFF
--- a/tensor2tensor/data_generators/audio.py
+++ b/tensor2tensor/data_generators/audio.py
@@ -60,7 +60,7 @@ def _collect_data(directory, input_ext, target_ext):
   #   if the datafile was "/path/to/datafile.wav" then the key would be
   #   "/path/to/datafile"
   # value: a pair of strings (input_filepath, target_filepath)
-  data_files = dict()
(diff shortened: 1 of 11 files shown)


afrozenator:
Thanks @lgeiger !

PR #1463, opened 2019-02-22
PPO mini-batches with tensors (instead of datasets)

piotrmilos:


lukaszkaiser:
This PR seems to make the tests fail and has some tricky indexing and integer division issues. Closing for now, let's re-open a new one once that is corrected.

PR #1468, opened 2019-02-25
Use xrange from six to fix Python 3 support

lgeiger:
T2T uses a mix of `xrange` and `range` leading to Python 3 errors when `xrange` isn't imported from `six.moves`.
This PR switches all usage of `xrange` to `six.moves.range` which is equivalent and supports both Python 3 and 2.


afrozenator:
Thanks again @lgeiger !

PR #1480, opened 2019-03-11
Allowing t2t-eval to use user-defined modules

wangcho2k:


lukaszkaiser:
Great thanks for making this PR, it will be very helpful to many users!

PR #1487, opened 2019-03-13
Fix step size extraction for checkpoint name 

lzhang10:
This fixes a bug introduced in https://github.com/tensorflow/tensor2tensor/commit/47ced4eecd13974e3a690a80b1f60effda089fa9

 Fix step size extraction for checkpoint name with - in it such as avg-model.ckpt-1234


lzhang10:
I signed it.

lzhang10:
I signed it!

afrozenator:
Thanks a lot for finding and fixing @lzhang10 !!

PR #1493, opened 2019-03-16
Change t2t-query-server into flask server.

Drunkar:
t2t-query-server was just a while input loop.
I changed it into a simple flask server.
I think this is easier to deploy a query server with docker or other environments.

### DIFF
--- a/tensor2tensor/serving/query.py
+++ b/tensor2tensor/serving/query.py
@@ -30,6 +30,10 @@ from tensor2tensor.utils import usr_dir
 from tensor2tensor.utils.hparam import HParams
 import tensorflow as tf
 
+from flask import Flask
+from flask import request
+from flask import jsonify
+
 flags = tf.flags
 FLAGS = flags.FLAGS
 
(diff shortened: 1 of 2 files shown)


Drunkar:
I signed it!

tensor2tensor/serving/query.py
```diff
@@ -85,21 +89,22 @@ def main(_):
...
-Output (Score {score:.3f}):
-{output}
-    """
-    print(print_str.format(inputs=inputs, output=output, score=score))
-    if FLAGS.inputs_once:
-      break
+
+  app = Flask(__name__.split(".")[0])
+
+  @app.route("/", methods=["GET"])
```
lukaszkaiser: This would be nice to have, but I don't think we can force everyone to use Flask. Could you change it so that Flask is just an option, maybe a parameter to the function that defaults to False (no Flask)?

tensor2tensor/serving/query.py
```diff
@@ -30,6 +30,10 @@
 from tensor2tensor.utils.hparam import HParams
 import tensorflow as tf
 
+from flask import Flask
```
lukaszkaiser: Let's make this import optional so not everyone needs to install it.

Drunkar:
@lukaszkaiser Thanks for your review!
It sounds a better way! will do it.

lukaszkaiser:
@Drunkar : are you still up for this? If not, I'll close in a bit and we can always re-open later.

Drunkar:
@lukaszkaiser Sorry I don't have times to do this. Please close this issue once.

lukaszkaiser:
No worries, we can always re-do. Closing for now.

PR #1555, opened 2019-04-23
extensibility for eval metrics

rllin-fathom:
new `Problem` property `all_metrics_fns` allows extending metrics functions in a `Problem` versus adding it to `metrics.METRICS_FNS`, especially if the function depends on problem information


afrozenator:
Thanks a lot @rllin-fathom !

PR #1577, opened 2019-05-20
Only strip end of line when loading the encoder from file.

funtion:
When loading from a file, only right side should be striped. Otherwise, when the file is unquoted and a token starts with space, the leading space will not be kept. This will lead to failure in line 627, because the original token could not be found.


funtion:
I signed it!

afrozenator:
Thanks @funtion !

PR #1578, opened 2019-05-20
Adding extra linear to semantic hashing discretization bottleneck.

martiansideofthemoon:
This fixes the semantic hashing bug mentioned here https://github.com/tensorflow/tensor2tensor/blob/master/tensor2tensor/models/research/transformer_vae.py#L870
@dustinvtran, @lukaszkaiser is this the correct way to fix this bug?

### DIFF
--- a/tensor2tensor/layers/discretization.py
+++ b/tensor2tensor/layers/discretization.py
@@ -309,6 +309,8 @@ def embed(x,
       h1a = tf.layers.dense(c, filter_size, name="vch1a")
       h1b = tf.layers.dense(1.0 - c, filter_size, name="vch1b")
       h1 = h1a + h1b
+      h1 = tf.layers.dense(h1, hidden_size, name="vch_final_linear")
+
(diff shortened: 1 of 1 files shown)


martiansideofthemoon:
I signed it!

afrozenator:
Hey @martiansideofthemoon -- thanks so much for the fix, I'll merge this in shortly, but would it be easy to add a unit test too to check if the fix works?

martiansideofthemoon:
Hi @afrozenator, what kind of a unit test would be useful? Should I compare the shapes of the input / output?

afrozenator:
Thanks a lot again for doing this @martiansideofthemoon ! Merging this now.

afrozenator:
@martiansideofthemoon -- 

> Hi @afrozenator, what kind of a unit test would be useful? Should I compare the shapes of the input / output?

Yes, anything to make sure that things run should be fine.

PR #1589, opened 2019-05-29
Documentation for creating own model

hbrylkowski:
I've written docs that were marked as TODO for a long time. 
Most of them are copied from appropriate places in the repository, but I think for someone who just starts with t2t it will be easier having it all in one place.
Also I've changed Contributions link to absolute one since it was 404'ing from docs directory.

### DIFF
--- a/docs/new_model.md
+++ b/docs/new_model.md
@@ -5,7 +5,7 @@ version](https://badge.fury.io/py/tensor2tensor.svg)](https://badge.fury.io/py/t
 [![GitHub
 Issues](https://img.shields.io/github/issues/tensorflow/tensor2tensor.svg)](https://github.com/tensorflow/tensor2tensor/issues)
 [![Contributions
-welcome](https://img.shields.io/badge/contributions-welcome-brightgreen.svg)](https://github.com/tensorflow/CONTRIBUTING.md)
+welcome](https://img.shields.io/badge/contributions-welcome-brightgreen.svg)](../CONTRIBUTING.md)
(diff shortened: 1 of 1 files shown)


docs/new_model.md
```diff
@@ -5,12 +5,103 @@ version](https://badge.fury.io/py/tensor2tensor.svg)](https://badge.fury.io/py/t
 [![GitHub
 Issues](https://img.shields.io/github/issues/tensorflow/tensor2tensor.svg)](https://github.com/tensorflow/tensor2tensor/issues)
 [![Contributions
-welcome](https://img.shields.io/badge/contributions-welcome-brightgreen.svg)](CONTRIBUTING.md)
+welcome](https://img.shields.io/badge/contributions-welcome-brightgreen.svg)](https://github.com/tensorflow/CONTRIBUTING.md)
```
martinpopel: This link does not exist. Use `../CONTRIBUTING.md` (or `https://github.com/tensorflow/tensor2tensor/blob/master/CONTRIBUTING.md` if absolute path is needed for some reason)

afrozenator:
This is really really awesome, thanks a LOT @hbrylkowski !

hbrylkowski:
@afrozenator - It's really good to hear that :) If I wanted to expand documentation further where would you say it is needed the most right now? I thought about documenting all models and datasets - extracting it from pydoc in model/dataset class, and then possibly add it to build so it would appear in docs automatically.

afrozenator:
Thanks again a lot for doing this @hbrylkowski , I'm adding @lukaszkaiser for more thoughts on this, but I'd say leave out datasets for now, since the long term plan is for that to live in https://www.tensorflow.org/datasets completely.

PR #1596, opened 2019-06-05
Using partial targets at inference time.

EugKar:
This pull request enables using `partial_targets` field of the inputs dictionary at inference time, when `inputs` field is also specified.
One of the cases when partial targets are useful is mutilingual speech recognition problem, as described in this paper: [Multilingual End-to-End Speech Recognition with A Single Transformer on Low-Resource Languages](https://arxiv.org/abs/1806.05059). Language tags may be specified at `partial_targets` input, if there is prior knowledge of the speaker's language.


EugKar:
I signed it!

afrozenator:
Thanks a lot @EugKar !!

Issue #917, opened 2018-07-05
t2t notebook attention visualization broken


tomsbergmanis:
### Description
When running https://colab.research.google.com/github/tensorflow/tensor2tensor/blob/master/tensor2tensor/notebooks/hello_t2t.ipynb
code for en2de translation and visualizing all attentions heads input-output view attentions are nonsensical:  for each output word all 8 attention heads to the first input word "The". Clearly, that is not right. 
The prerendered visualization shows a very different attentions that do not match those produced by code. 

I got this issue by simply running the notebook. 
![screenshot from 2018-07-05 18 02 55](https://user-images.githubusercontent.com/3306933/42337486-dbc64c00-807e-11e8-897c-422d2553119a.png)

martinpopel:
Can you try installing an older version of T2T in Colab (e.g. `pip install tensor2tensor==1.4.0`) and check if there is the same problem?
I have never seen this kind of problem, but I remember the visualization code was always buggy: for some sentences there were attention positions with no tokens shown or the other way round. I was just lazy to inspect it properly and report an issue.

Note that the enc-dec (input-output) attention may not correspond to what we would expect as alignment (it may be e.g. "shifted by one", cf. #898) - thus it is difficult to distinguish these unexpected attentions from visualization errors.

tomsbergmanis:
Thank you for your answer. 

After installing pip install tensor2tensor==1.4.0 code fails to import trainer_lib:
ImportErrorTraceback (most recent call last)
<ipython-input-2-d2b4f8fee4dd> in <module>()
      8 from tensor2tensor import problems
      9 from tensor2tensor.layers import common_layers
---> 10 from tensor2tensor.utils import trainer_lib
     11 from tensor2tensor.utils import t2t_model
     12 from tensor2tensor.utils import registry

ImportError: cannot import name trainer_lib

tomsbergmanis:
Any other suggestions?

repoloper:
+1

Not sure if related, but one problem I am seeing is that attention is not using the decode_hparams somehow and the translations are different when I use t2t-decoder vs. the one in the attention visualization.

drscotthawley:
I don't even get any display of anything.  No matter what I select for the dropdowns, there are no layers to select and no output from this portion of the notebook.

Issue #898, opened 2018-06-26
Can I get a word-alignment list through Transformer in nmt work?


dean248248:
### Description
I want to get a word-alignment list in nmt system, but I don't know how to do this. Can this framework achieve my goal?
...

### Environment information

```
OS: <your answer here>

$ pip freeze | grep tensor
# your output here

$ python -V
# your output here
```

### For bugs: reproduction and error logs

```
# Steps to reproduce:
...
```

```
# Error logs:
...
```

martinpopel:
See the [hello_t2t IPython notebook at Google Colab](https://colab.research.google.com/github/tensorflow/tensor2tensor/blob/master/tensor2tensor/notebooks/hello_t2t.ipynb#scrollTo=T7UJzFf6fmhp) and in the "Display Attention" section choose "Input - Output" instead of "All" from the menu.
This will visualize the encoder-decoder attention, which is the closest thing to word alignment you can get out of the box (as far as I know).
As you can see, this is very different from the word alignment you expected:

- The attention is fuzzy, you would need to threshold it (or choose max) to get binary alignment (aligned or not aligned).
- There are multiple attention heads and most of them seem to be unrelated to word alignment.
- It is well known the attention may be shifted by one (esp. in case of LSTMs, which is not the case here, so I am not sure)
- I am afraid the visualization code contains a bug as the there are attention positions with no tokens shown or the other way round (try entering a different sentence).
- The attention is between subwords, not words. More precisely, it is between positions corresponding to subwords, but enc-dec attention connects the deepest layer, where the vector representation on each position represents theoretically the whole sentence (with different emphasis on different parts), not just the "current" subword.

Of course, you could train a new T2T model in the "old NMT way": restrict the number of enc-dec attention heads to 1, use words (with UNK) instead of subwords,... and get much worse MT quality, but with a better interpretable enc-dec attention.

tomsbergmanis:
Using the same visualizer and the ipython notebook I get really bizarre attentions.
What could be the issue? 
![screenshot from 2018-07-05 18 02 55](https://user-images.githubusercontent.com/3306933/42337782-aaf1c036-807f-11e8-936b-e53ce481fbb0.png)

PR #1602, opened 2019-06-12
Storing encoder-decoder attention history at Transformer's cache during fast decoding.

EugKar:
Sometimes, attention weights may be useful for explaining model's results, visualization, etc. At speech recognition, encoder-decoder attention may be used for time alignment of waveform with recognized phonemes. In old releases, or without fast decoding, the attention is stored at `attention_weights`, but it has been broken at fast decoding mode: #898 ,  #917 .
This PR returns weights in `cache["attention_history"]`, works in both Eager and Graph modes.


lukaszkaiser:
Thanks for correcting this! I had to tweak a few things, but hopefully it all works now. Thanks again!

PR #1626, opened 2019-07-08
Add English-Spanish translation problem

voluntadpear:
Hi!

I was able to successfully add and use a English-Spanish translation problem, using **Common Crawl**, **EuroParl v7**, **UN Corpus** and **ParaCrawl** as my datasets.

I trained both an EN-ES and an ES-EN Transformer models (using the `transformer_big_single_gpu` hyperparameters set) with this new problem specification.

### DIFF
--- a/README.md
+++ b/README.md
@@ -210,6 +210,7 @@ There are a number of translation data-sets in T2T:
 * English-Czech: `--problem=translate_encs_wmt32k`
 * English-Chinese: `--problem=translate_enzh_wmt32k`
 * English-Vietnamese: `--problem=translate_envi_iwslt32k`
+* English-Spanish: `--problem=translate_enes_wmt32k`
 
 You can get translations in the other direction by appending `_rev` to
 the problem name, e.g., for German-English use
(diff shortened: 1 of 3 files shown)


voluntadpear:
Done CLI

lukaszkaiser:
Did you manage to get good results from the models? Great thanks for contributing!

iamyihwa:
Hello! Thanks for adding en-es module and making it available! 

I have ran into an issue when I tried to download data using 't2t-datagen' command. 

>export PROBLEM=translate_enes_wmt32k
> t2t-datagen --data_dir=$DATA_DIR --tmp_dir=$TMP_DIR --problem=$PROBLEM

It works when I use english to german module for example by doing  
>export PROBLEM=translate_ende_wmt32k



Some part of the error message that I get clearly show that spanish module is not part of the available one at the moment.  (there is no enes module appearing where it should be since things seem to be ordered alphabetically.)

``` 

ValueError: You must specify one of the supported problems to generate data for:
  * algorithmic_addition_binary40
...

  * translate_ende_wmt_clean_pc_clean32k
  * translate_ende_wmt_multi64k
  * translate_ende_wmt_multi64k_packed1k
  * translate_ende_wmt_pc32k
  * translate_ende_wmt_pc_clean32k
  * translate_enet_wmt32k
  * translate_enet_wmt_characters
  * translate_enfr_wmt32k
```

I have tried installing through 'pip install tensor2tensor' as well as 'pip install git+https://github.com/tensorflow/tensor2tensor.git' to check if the last changes were not reflected for the official release.

PR #1635, opened 2019-07-19
Model-Based RL: Rainbow params with larger epsilon.

konradczechowski:


afrozenator:
Thanks @konradczechowski !

PR #1645, opened 2019-07-25
fix decode bug

dong-s:
decode function used input vocab instead of target vocab.


dong-s:
I signed it

afrozenator:
Thanks a lot @dong-s for this!

PR #1651, opened 2019-07-28
correct typo in add_timing_signal_nd

Separius:


afrozenator:
Thanks a lot @Separius !

PR #1666, opened 2019-08-20
WIP: Added new initializers

joaogui1:
Adding kaiming uniform and normal and lecun uniform and normal initializers, besides adding the more general variance scaling initializer

### DIFF
--- a/tensor2tensor/trax/layers/initializers.py
+++ b/tensor2tensor/trax/layers/initializers.py
@@ -22,6 +22,18 @@ from __future__ import print_function
 import numpy as onp
 from tensor2tensor.trax import backend
 
+def _get_fans(shape):
+    receptive_field = np.prod(shape[:-2])
+    if len(shape) >= 2:
+        fan_in, fan_out = shape[-2], shape[-1]
+    elif len(shape) == 1:
+        fan_in, fan_out = shape[0]
+    else:
+        fan_in, fan_out = 1.
+    fan_in *= receptive_field
+    fan_out *= receptive_field
+    return fan_in, fan_out
+
 
 def RandomNormalInitializer(stddev=1e-2):
   """An initializer function for random normal coefficients."""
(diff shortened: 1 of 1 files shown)


joaogui1:
I believe I'm done here

tensor2tensor/trax/layers/initializers.py
```diff
@@ -22,6 +22,18 @@
 import numpy as onp
 from tensor2tensor.trax import backend
 
+def _get_fans(shape):
+    receptive_field = np.prod(shape[:-2])
+    if len(shape) >= 2:
+        fan_in, fan_out = shape[-2], shape[-1]
```
lukaszkaiser: In trax we want to have the control to specify which dimensions the fans are on -- that what in-dim and out-dim are used for. Please pass the arguments here and allow an option to use the whole receptive field too.

tensor2tensor/trax/layers/initializers.py
```diff
@@ -30,21 +42,70 @@ def Init(shape, rng):
...
-  def Init(shape, rng):
-    fan_in, fan_out = shape[in_dim], shape[out_dim]
-    std = backend.numpy.sqrt(2.0 / (fan_in + fan_out))
-    a = backend.numpy.sqrt(3.0) * std
-    return backend.random.uniform(rng, shape, minval=-a, maxval=a)
-  return Init
+def VarianceScalingInitializer(scale, mode, distribution):
+    """Initializer capable of adapting its scale to the shape of weights tensors."""
+    if scale <= 0.:
+        raise ValueError(f"scale must be positive float, {scale} given")
```
lukaszkaiser: We need both py2 and py3 comatibility, so please don't use f"aaa" -- just do 'aaa' instead. (And note that in trax we try to use single-quotes.)

tensor2tensor/trax/layers/initializers.py
```diff
@@ -30,21 +42,70 @@ def Init(shape, rng):
   return Init
 
 
-def GlorotNormalInitializer(out_dim=0, in_dim=1, scale=onp.sqrt(2)):
-  """An initializer function for random Glorot-scaled coefficients."""
+def RandomUniformInitializer(lim=1.0):
+  """An initializer function for random uniform coefficients."""
   def Init(shape, rng):
```
lukaszkaiser: Please use 2-spaces for indent everywhere (google style).

tensor2tensor/trax/layers/initializers.py
```diff
@@ -30,21 +42,70 @@ def Init(shape, rng):
...
+def KaimingNormalInitializer(out_dim=0, in_dim=1, param=0.):
+    """An initializer function for random Kaiming-scaled coefficients."""
+    return VarianceScalingInitializer(2.0 / backend.np.sqrt(1 + param**2),
+                                      "fan_in", "truncated_normal")
+
+
+def KaimingUniformInitializer(out_dim=0, in_dim=1, param=0.):
+    """An initializer function for random uniform Kaiming-scaled coefficients."""
+    return VarianceScalingInitializer((2.0 / backend.np.sqrt(1 + param**2),
+                                      "fan_in", "uniform")
```
lukaszkaiser: Brackets are missing here at the end.

joaogui1:
Implemented the requested changes

tensor2tensor/trax/layers/initializers.py
```diff
@@ -30,21 +41,70 @@ def Init(shape, rng):
...
+    elif mode == 'fan_out':
+      gain /= fan_out
+    elif mode == 'fan_avg':
+      gain /= (fan_in + fan_out) / 2
+    if distribution == 'truncated_normal':
+      # constant from scipy.stats.truncnorm.std(a=-2, b=2, loc=0., scale=1.)
+      stddev = backend.numpy.sqrt(gain) / .87962566103423978
+      return (backend.random.truncated_normal(rng, -2, 2, shape) * stddev).astype('float32')
+    elif distribution == 'normal':
+      return (backend.random.normal(rng, shape, dtype) * backend.numpy.sqrt(gain)).astype('float32')
```
lukaszkaiser: dtype is an udefined variable here

tensor2tensor/trax/layers/initializers.py
```diff
@@ -30,21 +41,70 @@ def Init(shape, rng):
...
   def Init(shape, rng):
-    fan_in, fan_out = shape[in_dim], shape[out_dim]
-    size = onp.prod(onp.delete(shape, [in_dim, out_dim]))
-    std = scale / backend.numpy.sqrt((fan_in + fan_out) / 2. * size)
-    return (std * backend.random.normal(rng, shape)).astype('float32')
+    return (backend.random.uniform(rng, shape, -lim, lim)).astype('float32')
   return Init
 
 
-def GlorotUniformInitializer(out_dim=0, in_dim=1):
```
lukaszkaiser: Shouldn't the default out_dim and in_dim be -1 and -2 as above in all initializers?

tensor2tensor/trax/layers/initializers.py
```diff
@@ -30,21 +41,70 @@ def Init(shape, rng):
...
+
+def KaimingNormalInitializer(out_dim=0, in_dim=1, param=0.):
+  """An initializer function for random Kaiming-scaled coefficients."""
+  return VarianceScalingInitializer(out_dim, in_dim, 2.0 / backend.np.sqrt(1 + param**2),
+                                    'fan_in', 'truncated_normal')
+
+
+def KaimingUniformInitializer(out_dim=0, in_dim=1, param=0.):
+  """An initializer function for random uniform Kaiming-scaled coefficients."""
+  return VarianceScalingInitializer((out_dim, in_dim, 2.0 / backend.np.sqrt(1 + param**2),
```
lukaszkaiser: This has double brackets, so it's not correct python?

joaogui1: Yes, weird how those tests were passing, thanks for catching it!

tensor2tensor/trax/layers/initializers.py
joaogui1: Yeah, those were from a older version of the code before I followed the standard of the other initializer that cast everything to float32, again weird that no test caught it. Fixed

tensor2tensor/trax/layers/initializers.py
joaogui1: Yes, you're correct again, I used what was written in the main branch without thinking. It is fixed now

PR #1669, opened 2019-08-21
Change confusing function name

lazylife7157:
Maybe _is_not_finished is more proper name for cond function because tf.while_loop keeps running while cond is True.


lazylife7157:
@googlebot I signed it!

lukaszkaiser:
You're right, this is much better! Thank you!

PR #1674, opened 2019-08-23
Update generator_utils.py

TanguyUrvoy:
Hi,
`isinstance(v[0], six.integer_types)` is False for `np.int64` type causing algorithmic_sort_problem data generation to fail.


TanguyUrvoy:
@googlebot I signed it!

PR #1677, opened 2019-08-26
achieve the paper Attention Is All You Need's position encoding

wuutiing:
as the paper Attention Is All You Need (https://arxiv.org/pdf/1706.03762.pdf) described, the position encoding was using sine and cosine function alternatively depending on the position's odevity. but the achievement in tensor2tensor just concated the sine and cosine which i think not right.


wuutiing:
@googlebot I signed it!

wuutiing:
> see #177 and read the discussion
> I would suggest to keep the code as is, just add there an explanatory comment to prevent future PRs on this issue.

sorry i have not notice the issue talked before, thank you for reminding. As is fully talked and decided , i agree with you and later @afrozenator . And i'm not a native english speaker, i think it's better you commit the comment change in code. thanks!

PR #1684, opened 2019-08-28
Activations

joaogui1:
OK, so I messed up a little bit and it seems the first few commits are from the other branch.
But anyway the relevant ones are from 28th August. 
Here is the colab justifying changing relu from maximum to clip:
https://colab.research.google.com/drive/1WjGacZNCGPUN-YzQiG9LNRCv49BwJJ4N


joaogui1:
Can someone Review this PR please?

PR #1689, opened 2019-09-03
Add missing input_dtype parameter to trax demo.

jlebar:


jlebar:
...actually it looks like this fix only applies if one tries to use a newer version of trax.

Do you all want the notebook to track the latest version?

lukaszkaiser:
I think we meant this notebook to stick to the one fixed version of Trax. We are starting to make new ones (see layers/demo ipynb) that are at-head. But there's some infrastructure (tests!) needed for that, so it took more time.

How about we leave this notebook as-is for now and migrate it all later? Should we just close this then?

jlebar:
> How about we leave this notebook as-is for now and migrate it all later? Should we just close this then?

SGTM, especially if there will be other demos at head.  I'm finding them really helpful.

PR #1690, opened 2019-09-07
WIP: Learning Rate Schedules

joaogui1:
Already added Polynomial and Exponential Decay. Over the weekend gonna try to add all of Keras learning rate schedules


joaogui1:
So, right now I've implemented everything, but I need some help/feedback with two things:

1. I think my use of random is kinda wrong in the noisy linear cosine decay, can someone help me fix it?
2. The code is really big now, so maybe I should write some tests for it, what do you guys think? If you think it's a good idea do you have any suggestions of how should I test the schedules?
Thank you for your time!

lukaszkaiser:
Sorry for the delay related to ICLR. Could you sync with the current head?

joaogui1:
@lukaszkaiser I believe the merge worked, it was very messy hahaha

lukaszkaiser:
Still seems to not run due to some (possibly new) changes...

joaogui1:
Oh, so that's the best practice when using JAX, but indeeed if the backend used is not jax or step is still not a DeviceArray this problem will appear, I will check len before choosing what to call them.

joaogui1:
Ok, I attempted a fix right now

lukaszkaiser:
This code does not work. It uses undefined variables and does not pass even a basic MNIST training, not to mention not containing tests. I'll close this for now -- please send a new PR containing tests and make sure by hand that training basic MNIST models (config/mlp_mnist.gin) runs with the new schedules. Furthermore, please run python linter before submitting for review, so basic code style is checked. Thanks!

PR #1695, opened 2019-09-08
[trax] Improve handling of data_dir.

jlebar:
- Give it a default.
 - Let it be specified on the command line unconditionally, rather than
   silently ignoring it sometimes.
 - Remove data_dir = None from model configs now that data_dir has a default.

### DIFF
--- a/tensor2tensor/trax/README.md
+++ b/tensor2tensor/trax/README.md
@@ -42,7 +42,7 @@ python -m tensor2tensor.trax.trainer \
 
 ```
 python -m tensor2tensor.trax.trainer \
-  --config_file=$PWD/trax/configs/resnet50_imagenet_8gb.gin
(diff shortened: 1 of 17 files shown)


jlebar:
cc @afrozenator

afrozenator:
@jlebar this is a little more subtle (see the internal code), but I believe Anselm's change should fix the issue you see here -- https://github.com/tensorflow/tensor2tensor/commit/9db59c5cba0c847b0e4e76d174f19147e044a0c1

Long story short, the default internally should be placer, which we do now if we find data_dir is set to None, externally Anselm's CL will kick in and give it a reasonable default.

Let me know if that works - I believe it should, unless I'm missing something? We can chat internally as well.

@lukaszkaiser

lukaszkaiser:
I think this became obsolete, closing for now. Please reopen if the issue is still there!

PR #1716, opened 2019-09-30
Return attn in t2t-model

steins024:
## What:
Return the input-output attention value of last layer (final decode layer dot attention value) in t2t-model

## Why:
Some task needs the attention infomation to predict alignment

## How(Changes):
1. return the attention value of last layer from attention_history in t2t_model.py
2. update decoding.py to support alignment in CLI
3. update serving/* to support alignment in tensorflow-serving


lukaszkaiser:
Sorry, this change is too invasive as it modifies core platform for the specific purpose of attention. We'd be happy to take in parts of it, like extra input in serving, but such prevalent modifications in core are too extensive to maintain.

PR #1733, opened 2019-10-31
En-Id untokenized parallel corpora

prasastoadi:
Change to untokenized parallel corpora due to t2t automatically tokenizes sentences.


afrozenator:
Thanks @prasastoadi !

PR #1748, opened 2019-11-20
Create Text2RealProblem class for regression-based problems

gabegrand:
A variety of NLP problems involve training regression models to predict continuous-valued ratings from text (e.g., sentiment scoring). While T2T supports several `REAL` modalities, there is no base class for regression-based text problems. This PR introduces a `Text2RealProblem` that is analogous to the existing `Text2ClassProblem`, but for regression problems.

The class also supports multitask regression via the `ntasks` property, which allows for the regression target to be an n-dimensional float vector.

Note that this PR doesn't contribute any additional problems to the registry (for now), since the datasets I've been training on are not public. That said, this base class makes it very straightforward to define a custom text regression problem, and I hope that it will be useful for other T2T users.

@lukaszkaiser This PR is very similar to the Gist you reviewed from me -- please let me know what you think!

### DIFF
--- a/tensor2tensor/data_generators/text_problems.py
+++ b/tensor2tensor/data_generators/text_problems.py
@@ -694,7 +694,6 @@ class Text2RealProblem(Text2TextProblem):
     return metrics_list
 
 
-
 def txt_line_iterator(txt_path):
   """Iterate through lines of file."""
   with tf.gfile.Open(txt_path) as f:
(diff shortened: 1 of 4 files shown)


afrozenator:
Thanks @gabegrand !!

gabegrand:
Much appreciated @afrozenator!

PR #1751, opened 2019-11-24
[UPDATE] fork

ruohoruotsi:
so we can do a head-to-head vs OpenNMT-py transformers for ADR & NMT tasks


ruohoruotsi:
Oops my PR is going in the wrong direction (should be from master to my fork), Sorry about the noise

PR #1752, opened 2019-11-24
Fix the decode_length for decoding

Michaelvll:
Originally the decode length is hard-coded in the `decoding.py`. 

I believe that is a bug because the `extra_length` set in `decode_hp` should be used as the `decode_length` rather than the original `tf.shape(x)[1] + 50`, referring to the usage of `features['decode_length']` in the followings:
https://github.com/tensorflow/tensor2tensor/blob/8107e53e8d8896c521522e9335a266b4b7510c37/tensor2tensor/utils/t2t_model.py#L932-L933 https://github.com/tensorflow/tensor2tensor/blob/8107e53e8d8896c521522e9335a266b4b7510c37/tensor2tensor/models/transformer.py#L712-L713

### DIFF
--- a/tensor2tensor/utils/decoding.py
+++ b/tensor2tensor/utils/decoding.py
@@ -453,7 +453,7 @@ def decode_from_file(estimator,
           task_id=decode_hp.multiproblem_task_id, has_input=has_input)
       gen_fn = make_input_fn_from_generator(input_gen)
       example = gen_fn()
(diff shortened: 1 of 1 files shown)


Michaelvll:
@googlebot I signed it!

afrozenator:
Thanks @Michaelvll !

PR #1760, opened 2019-12-01
Fixed a mistake in a doc comment

Hamza5:
In T2TModel.infer, the comment mentioned that the returned outputs tensor has a shape of `[batch_size, <= decode_length] if beam_size == 1`, but this is wrong. It should be: `if top_beams == 1`. It is a simple mistake but it may really confuse a beginner (like me).


Hamza5:
@googlebot I signed it!

afrozenator:
Thanks @Hamza5 !

PR #1766, opened 2019-12-09
Add support for training on V100 GPUs on Cloud ML

thominj:
Updated the runtime version to 1.14 and added support for
using V100 GPUs on Cloud ML.

Test plan:
I ran several jobs with a custom problem definition on Cloud ML using each machine type (P100 and V100, medium and large) to make sure they worked.


afrozenator:
Thanks @thominj !

PR #1781, opened 2020-01-09
proposed fix for issue 1461

josd:


josd:
@googlebot I signed it!

PR #1784, opened 2020-01-29
Inference support for Neural Shuffle-Exchange network

OptimatiumFeles:
When target dimensions are not known, use hparams.max_length. max_length should be a power of 2.
Max length can be enforced using hparams.force_max_length=True.
Auto-Regressive decoding not supported at this moment.

### DIFF
--- a/tensor2tensor/models/research/shuffle_network.py
+++ b/tensor2tensor/models/research/shuffle_network.py
@@ -420,7 +420,7 @@ class ShuffleNetwork(t2t_model.T2TModel):
     # Run the model
     self.hparams.force_full_predict = True
     with tf.variable_scope(self.name):
-      logits, losses = self.model_fn(features)
+      logits, _ = self.model_fn(features)
 
     assert len(logits.shape) == 5  # [batch, time, 1, 1, vocab]
     logits = tf.squeeze(logits, [2, 3])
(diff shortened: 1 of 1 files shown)


dfurrer:
I get the following linter error (and a crash when trying to run). Could you fix this in the request please?

`
third_party/py/tensor2tensor/models/research/shuffle_network.py:
364: Type annotations not enabled. (<string>, line 364) [syntax-error]
`

OptimatiumFeles:
I hope this solves the problem.

dfurrer:
Thanks. I have it running and the decode calls seem to work. However the linter now complains about these:
347: There should not be a space before a one-line docstring summary [g-space-before-docstring-summary]
347: One-line docstring summary should be followed by a blank line [g-no-space-after-docstring-summary]
389: There should not be a space before a one-line docstring summary [g-space-before-docstring-summary]
389: One-line docstring summary should be followed by a blank line [g-no-space-after-docstring-summary]
389: Missing Returns section in docstring, or wrong choice of Returns/Yields [g-doc-return-or-yield]

OptimatiumFeles:
Fixed those errors as well.

afrozenator:
Many thanks to @OptimatiumFeles and @dfurrer both!

PR #1805, opened 2020-04-21
Residual Shuffle-Exchange network

EmilsOzolins:
T2T  implementation of the Residual Shuffle-Exchange network.

Paper: https://arxiv.org/abs/2004.04662
Original code: https://github.com/LUMII-Syslab/RSE


afrozenator:
Thanks @EmilsOzolins !

PR #1834, opened 2020-07-17
add a backtranslation-based en-cs NMT problem and hparams

martinpopel:
This code should serve as a reference (to be cited in a paper),
so I would prefer to have it in a (single) separate file.


lukaszkaiser:
Sorry Martin - we cannot import transformer in data_generators. We could copy all hparams, or add a comment. For now merging and it'll be moved.

PR #1852, opened 2020-09-17
Fix CommonVoice dataset for Speech Recognition problem

RegaliaXYZ:
Fixed Common Voice data generator by adding a flag to the `datagen.py` file for the language code (`--language="en"`, if not specified it defaults to english) and dynamically downloading the correct language dataset.

Also had to rework the architecture of the data unpacking since Mozilla changed their folder architecture.

Also removed the sub-problems of Common Voice (Noisy, Clean, FullTestClean since all the previous .tsv files were merged (no more other-train, other-test etc)


RegaliaXYZ:
@googlebot I signed it!

PR #1860, opened 2020-10-22
Resolving subdependencies conflict

yw662:
kfac > 0.2 requires tensorflow-probability==0.8.0. However, T2T requires tf-probability==0.7.0.
Restricting kfac to ==0.1.5 so that it can work.


yw662:
@googlebot I signed it!

PR #1890, opened 2021-06-27
Remove extra `.value` in `tensor.shape[-1].value`

paulo-raca:
I couldn't figure out why the `.value` was ever added, but running it results in `AttributeError: 'int' object has no attribute 'value'`


paulo-raca:
Abandoning

Issue #1911, opened 2022-05-26
AttributeError: 'DummyModule' object has no attribute 'load_checkpoint'


almostimplemented:
### Description
In the TensorFlow v2 codepath, there is a bug preventing loading the checkpoint. 

The bug is very clear in the code. In `tensor2tensor/utils/contrib.py`, in the absence of `tensorflow.contrib`, the method `framework()` will return a `DummyModule`. Then, in `tensor2tensor/utils/t2t_model.py`, we try to load the checkpoint via:

```
reader = contrib.framework().load_checkpoint(ckpt_dir)
variable_map = {}
for var in contrib.framework().get_trainable_variables():
```

I will open a pull request with my personal solution, and I am open to change it to best fit the project.

...

### Environment information

```
OS: Linux

(venv) ace01@wynne:~/forks/magenta$ pip freeze | grep tensor
mesh-tensorflow==0.1.21
-e git+https://github.com/tensorflow/tensor2tensor@c8fe559e0b357389d8754474e1306b6ca9afc4f3#egg=tensor2tensor
tensorboard==2.9.0
tensorboard-data-server==0.6.1
tensorboard-plugin-wit==1.8.1
tensorflow==2.9.1
tensorflow-addons==0.17.0
tensorflow-datasets==4.5.2
tensorflow-estimator==2.9.0
tensorflow-gan==2.1.0
tensorflow-hub==0.12.0
tensorflow-io-gcs-filesystem==0.26.0
tensorflow-metadata==1.8.0
tensorflow-probability==0.7.0


(venv) ace01@wynne:~/forks/magenta$ python -V
Python 3.7.7
```

### Reproduction notes

This is not a minimal reproduction -- this is simply how I encountered it. I am finetuning a [musical note sequence model from Magenta](https://github.com/magenta/magenta/tree/main/magenta/models/score2perf).

I think the bug is obvious in the code, so I do not feel the need to provide a minimal repro, but I can if it is deemed necessary. 

```
(venv) ace01@wynne:~/forks/magenta$ cat finetune_midi_transformer.sh
DATA_DIR=/homes/ace01/forks/magenta/jazz_piano_datagen
HPARAMS_SET=transformer_tpu
MODEL=transformer
PROBLEM=score2perf_jazz_piano_language_uncropped_aug
TRAIN_DIR=/homes/ace01/forks/magenta/finetune_dir
[...]

PR #1912, opened 2022-05-26
Fix resume checkpoint for tf2 codepath

almostimplemented:
Cheap, localized solution to the bug described in #1911 

For the TF2 codepath, provide additional two methods to the `DummyModule`: `load_checkpoint` and `get_trainable_variables`.


almostimplemented:
Hm, maybe these can be added directly to the `DummyModule` construction, down at `def framework` (line 83 in original).

